PR #4, opened 2015-08-30
Freeze panes functionality

pedromarins:
Adding freeze panes functionality to patient chart table.

Move the first row so if you scroll vertically it follows the scroll.

Move the first column so if you scroll horizontally it follows the scroll.


zestyping:
Thanks for contributing this implementation!  We have a frozen headers implementation that seems to work pretty well for now, so I'm closing this pull request.

PR #5, opened 2015-08-31
Grid fixed headers a.k.a. Freeze Panes

llvasconcellos:
Added fixed header and column on the patient's history grid.


zestyping:
Thanks for contributing this implementation!  We have a frozen headers implementation that seems to work pretty well for now, so I'm closing this pull request.

PR #11, opened 2015-09-21
Adding set up instructions to tablet detection

viniciusboson:
Adding set up instructions to tablet detection


zestyping:
Thank you for adding these instructions!

PR #12, opened 2015-09-22
Fixing compilation errors in LocalDate/DateTime convertion

viniciusboson:
The birthdate field is already `LocalDate` and doesn't need to be converted from `DateTime`.


zestyping:
Closing this change as its the same as #14.  Thank you for fixing it!

PR #35, opened 2015-09-25
Add ObsValue and ObsPoint; provide JSON data to support custom JS rendering in the chart.

zestyping:
This separates the ideas of "observations", "points", and "values".
- A value is an observed quantity, such as a number, a coded value, or a piece of text.
- A point is a value together with the time it was observed (the name "point" is intended to be suggestive of a point in a time-series graph).
- An observation specifies which concept was observed, together with a point.

These are now separate classes, for type-safety and efficiency.
- ObsValue is an implementation of a type union that represents values.
- ObsPoint is an ObsValue with an Instant.
- Obs **should** be a concept UUID with an ObsPoint, but I held off from making that big change just yet.

This refactoring was done to support the exporting of data to JSON so that custom JS functions can render it in the chart: what we are exporting for a given concept (e.g. weight) is an array of ObsPoints to be rendered.  It also significantly improves the robustness of ObsFormat and the associated customizable formatting machinery.

The JS API is as follows:
- For rendering a value in a tile, you define a function(ids, series1, series2, ...) where "ids" is an array of the concept IDs (remember you can specify more than one in the "concept" column even though usually it will be just one), and series1 is an array of points.  Each point is represented by a JSON object with a "time" field and one other typed field, e.g. {"time": _milliseconds_, "number": _numeric_value_} for numeric points.
- For rendering a value in a grid row, you define a function(ids, groups1, groups2, ...) where "ids" is an array of the concept IDs (remember you can specify more than one in the "concept" column even though usually it will be just one), and groups1 is an array of "point groups".  Instead of receiving one big array containing all the points, your function receives the points grouped according to the table cell they belong to.  Each point group is represented by a JSON object with an array of points in its "points" field, a reference to the DOM table cell element in the "cell" field, and the cell's time interval given in "start" and "stop" fields (in milliseconds since 1970-01-01 00:00 UTC).  This way you can choose to render the data into the appropriate cells or position any rendered graphics with respect to the cells in which the data belongs.


zestyping:
This update cleans up the API a little bit (functions receive the DOM cell directly instead of having to construct the element ID and find it themselves).

app/src/main/java/org/projectbuendia/client/models/Obs.java
```diff
@@ -28,62 +27,77 @@
...
-public final class ObsValue implements Comparable<ObsValue> {
-    public static final Comparator<ObsValue> BY_OBS_TIME = new Comparator<ObsValue>() {
-        @Override public int compare(ObsValue left, ObsValue right) {
-            return left.obsTime.compareTo(right.obsTime);
-        }
-    };
-
+// TODO: Make ObsPoint a member of Obs; change the structure of Obs to be simply:
+// { final @Nonnull String uuid; String name; final @Nonnull ObsPoint point; } then delete
+// getObsPoint(), getObsValue(), compareTo(), getTypeOrdering(), getCodedValueOrdering().
```
dancunningham, line 32: compareTo() for ObsPoint is based on time, whereas compareTo() for Obs has some more sophisticated rules including hard-coded ordering by severity for getCodedValueOrdering(). I don't particularly like those hard-coded rules, but it seems like we would have to find another way of ordering by severity if we dropped these methods.

dancunningham, line 32: I see that ObsValue on the other hand does have the logic to compare coded values.

Should ObsPoint compare by time (as it does now) or by value (which would use the ObsValue compareTo() method)?

app/src/main/java/org/projectbuendia/client/models/ObsValue.java
```diff
@@ -0,0 +1,247 @@
...
+    private static final Map<String, Integer> CODED_VALUE_ORDERING =
+        new ImmutableMap.Builder<String, Integer>()
+            .put(ConceptUuids.NO_UUID, -100)
+            .put(ConceptUuids.NONE_UUID, -1)
+            .put(ConceptUuids.NORMAL_UUID, -1)
+            .put(ConceptUuids.SOLID_FOOD_UUID, -1)
+            .put(ConceptUuids.MILD_UUID, 1)
+            .put(ConceptUuids.MODERATE_UUID, 2)
+            .put(ConceptUuids.SEVERE_UUID, 3)
+            .put(ConceptUuids.YES_UUID, 100).build();
```
dancunningham, line 73: Should we add others in here where they do have a severity ordering, like AVPU values and Oedema?

I don't like that these are hard-coded though. How might we have the ordering defined in the CSV import? What if the convention was to always order from low severity to high severity in the IDs used?

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -116,62 +127,100 @@ public void render(Chart chart, Map<String, ObsValue> latestObservations,
...
+                    ObsPoint[] points = new ObsPoint[item.conceptUuids.length];
+                    for (int i = 0; i < points.length; i++) {
+                        Obs obs = latestObservations.get(item.conceptUuids[i]);
+                        if (obs != null) {
+                            points[i] = obs.getObsPoint();
+                        }
+                    }
+                    tileRow.add(new Tile(item, points));
+                    if (!item.script.trim().isEmpty()) {
+                        mConceptsToDump.addAll(Arrays.asList(item.conceptUuids));
```
dancunningham, line 139: Currently all tile and row concepts are dumped to JSON right? Should we limit it to dump only those with something in the "script" field? Would there be any reason to dump a concept to JSON if there is nothing in the script field?

dancunningham, line 139: Oops - just read the line above; I see that is what it currently is doing. Great!

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -116,62 +127,100 @@ public void render(Chart chart, Map<String, ObsValue> latestObservations,
...
+                        JSONArray pointArray = new JSONArray();
+                        SortedSet<ObsPoint> points = column.pointSetByConceptUuid.get(uuid);
+                        if (points != null && points.size() > 0) {
+                            for (ObsPoint point : points) {
+                                pointArray.put(point.toJson());
+                            }
+                            JSONObject pointGroup = new JSONObject();
+                            pointGroup.put("start", column.start.getMillis());
+                            pointGroup.put("stop", column.stop.getMillis());
+                            pointGroup.put("points", pointArray);
```
dancunningham, line 214: Would also be really handy to have min and max values to set a sensible range for the y axis.

How might we do target weight as well?

dancunningham:
I am finding that the chart loads very slowly now that we have 25+ days worth of columns for some patients.

In addition, when it initially renders the web view with the grid, there is a delay and it is frozen until the weight graph has finished drawing.

app/src/main/assets/chart.html
```diff
@@ -2,11 +2,7 @@
 <link rel="stylesheet" href="chart.css">
```
llvasconcellos, line 2: The chart.html is missing the following tags:
&lt;!DOCTYPE html&gt;
&lt;html&gt;
&lt;meta charset="UTF-8"&gt;

This is good to assure the proper rendering of the page.

app/src/main/assets/chart.html
```diff
@@ -15,12 +11,13 @@
     <tr>
       {% for tile in tileRow %}
```
llvasconcellos, line 12: Well, since I'm here to comment, I'm really not a fan of this template engine. We should have natural html files (that can be read by the browser even without the data) with clearer code. Html partials to help organization. Complete separation of the view from the logic. I think this is very messy and today we have better ways to generate dynamic html. What are your thoughts on using javascript for this? Don't get me wrong. The way it is is perfectly fine. It works. I just think there are better ways of organizing and generating html.
Java classes could just generate JSON and let webview html + javascript do the rest.

app/src/main/java/org/projectbuendia/client/models/ObsValue.java
zestyping, line 73: Yes, I agree it would be better to define the ordering elsewhere and not have it hardcoded here.  At the moment this seems the only place to put this logic, though.  

To get it from the profile would require:
- Defining how it should be specified in the CSV profile
- Deciding where to store this information in OpenMRS, and writing the code (in buendia-profile-apply) to import it into OpenMRS
- Deciding how to expose this information via the API, and extending the Buendia OpenMRS module to provide the ordering information through the API
- Deciding where/how to store this information on the client, and writing the client code to sync the information down from the server to the client
- Modifying the client to use this information when sorting ObsValues.

So, it's a substantial piece of work.  I suggest we leave a TODO comment here to look into this possibility in the future and proceed with the hardcoded implementation for now.

app/src/main/java/org/projectbuendia/client/models/Obs.java
zestyping, line 32: Yes, it's intentional that ObsPoint compares by time whereas ObsValue compares by value/severity.  Both kinds of sorting are useful, and these seem like reasonable defaults to me, especially when you think of it as:
- ObsPoint = (x, y), which is sorted by x
- ObsValue = y, which is sorted by y

I agree it would be good to clarify, though.  If anyone wants to follow up after this lands by editing the JavaDoc descriptions of these classes to mention the sort order, that sounds good to me.

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
zestyping, line 139: Ok good!

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
zestyping, line 214: Yes, we could add min and max as well.  Maybe in a follow-up change, after we land this one?  We can always add more properties to the JSON object.

Target weight is a synthetic concept.  We need to compute the target weight from the other observations by defining a JS function.  See the Trello card for my thoughts on a proposed design for that: https://trello.com/c/mEfdKOC2/44-synthetic-concepts

app/src/main/assets/chart.html
zestyping, line 12: I worry that this would be significantly slower with JavaScript, and would involve writing quite a bit more code than we have here.  Do you think so?  If you like, you could try experimenting with a simple example, and seeing how much JS you need to write to get it to work and how fast it is.  If it is at least as fast and can be expressed in code that isn't longer or more complex than the template, I'd support that change.

Again, however, I should caution that this isn't a priority right now — we don't need to redesign how this works for the field trip, and time is better spent on other needs for the field trip, not on this.  It could be a longer-term design question to return to after the trip, when we have time for some experimentation and further analysis.

zestyping:
Bugs #46 and #47 should both be fixed now at the tip of this branch.

zestyping:
So this should be ready to try out (give it a quick run through on the tablet, using nutrition-dan14-with-graph.csv as the profile on the server, to confirm the bugs are fixed) and merge to dev.

nutrition-dan14-with-graph.csv should be safe to apply in any case, whether or not this is ultimately the version we go with — it only adds a script to draw the bar graph, which is ignored by previous versions of the client.

dancunningham:
@zestyping the reason I rolled back nutrition-dan14-with-graph.csv is that on the dev branch it broke the alignment of the chart with a grey space below the weight row.

Thanks for fixing the bugs, I will test this Tuesday morning and close the bugs once I verify they are fixed.

zestyping:
Oh, I think I know what you mean.  The weight row is made taller, and the
frozen header is not expanded to match, right?  This is fixed in the "adjust styles" PR -- the frozen headers are automatically resized to match.

llvasconcellos:
@dancunningham, how do I test this?

> @zestyping the reason I rolled back nutrition-dan14-with-graph.csv is that on the dev branch it broke the alignment of the chart with a grey space below the weight row.
> 
> Thanks for fixing the bugs, I will test this Tuesday morning and close the bugs once I verify they are fixed.

app/src/main/assets/chart.html
llvasconcellos, line 12: OK, I know this is a long therm thing. I just wanted to know your thoughts on this. 
I'll do a test implementation at the proper time.
What about the <html> tag I mentioned here: https://github.com/projectbuendia/client/pull/35#discussion_r40755743 ?

app/src/main/assets/chart.html
llvasconcellos, line 2: To add more considerations on the matter of the <!DOCTYPE html> and <html> tags here is a screenshot I made:
screenshot from 2015-10-06 01 58 14

During the implementation of the fixed headers I noticed that the html output of pebble did not add the html and doctype tags.
As you can see, the browser added the html tag even though it wasn't present in the html file. This doesn't mean that we shouldn't use the html tag. 
Explicitly declaring html5 document would avoid odd behaviors when using most advanced html features. This is to tell exactly how the browser (webview) should render the html.
Here are some info:
http://www.w3.org/QA/2002/04/valid-dtd-list.html
http://www.w3.org/wiki/Choosing_the_right_doctype_for_your_HTML_documents
https://hsivonen.fi/doctype/
http://alistapart.com/article/doctype
http://www.w3schools.com/tags/tag_doctype.asp

dancunningham:
@zestyping yes that's exactly what I mean, though it's not an issue once we merge this PR.

app/src/main/java/org/projectbuendia/client/models/ObsValue.java
llvasconcellos, line 73: Added a trello board card for this.

PR #39, opened 2015-09-30
Enable Stetho to intercept and inspect network requests.

zestyping:
Stetho can show you all the network requests made by our app, in the Network tab, in Chrome Developer Tools.  But, to make that work, we have to install a StethoInterceptor to intercept our Volley requests, and that requires an OkHttp-based network stack.  This change adds an open source OkHttpStack class by Bryan Stern (https://gist.github.com/bryanstern/4e8f1cb5a8e14c202750) and installs the interceptor.

image

### DIFF
--- a/app/src/main/java/com/circle/android/api/OkHttpStack.java
+++ b/app/src/main/java/com/circle/android/api/OkHttpStack.java
@@ -48,8 +48,6 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import javax.annotation.Nullable;
-
 
 /**
  * OkHttp backed {@link com.android.volley.toolbox.HttpStack HttpStack} that does not
@@ -65,7 +63,7 @@ public class OkHttpStack implements HttpStack {
 
     @Override
     public HttpResponse performRequest(Request<?> request, Map<String, String> additionalHeaders)
-        throws IOException, AuthFailureError {
+            throws IOException, AuthFailureError {
 
         OkHttpClient client = mClient.clone();
         int timeoutMs = request.getTimeoutMs();
@@ -121,7 +119,7 @@ public class OkHttpStack implements HttpStack {
 
     @SuppressWarnings("deprecation")
     private static void setConnectionParametersForRequest(com.squareup.okhttp.Request.Builder builder, Request<?> request)
-        throws IOException, AuthFailureError {
+            throws IOException, AuthFailureError {
         switch (request.getMethod()) {
             case Request.Method.DEPRECATED_GET_OR_POST:
                 // Ensure backwards compatibility.  Volley assumes a request with a null body is a GET.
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/circle/android/api/OkHttpStack.java
```diff
@@ -0,0 +1,183 @@
...
+        entity.setContentEncoding(r.header("Content-Encoding"));
+
+        if (body.contentType() != null) {
+            entity.setContentType(body.contentType().type());
+        }
+        return entity;
+    }
+
+    @SuppressWarnings("deprecation")
+    private static void setConnectionParametersForRequest(com.squareup.okhttp.Request.Builder builder, Request<?> request)
```
viniciusboson, line 121: You should break the line here, it has more than 100 characters.

app/src/main/java/com/circle/android/api/OkHttpStack.java
```diff
@@ -0,0 +1,183 @@
...
+        throws IOException, AuthFailureError {
+        switch (request.getMethod()) {
+            case Request.Method.DEPRECATED_GET_OR_POST:
+                // Ensure backwards compatibility.  Volley assumes a request with a null body is a GET.
+                byte[] postBody = request.getPostBody();
+                if (postBody != null) {
+                    builder.post(RequestBody.create(MediaType.parse(request.getPostBodyContentType()), postBody));
```
viniciusboson, line 128: This line has more than 100 characters. It should be indented.

app/src/main/java/com/circle/android/api/OkHttpStack.java
```diff
@@ -0,0 +1,183 @@
...
+            okHttpRequestBuilder.addHeader(name, additionalHeaders.get(name));
+        }
+
+        setConnectionParametersForRequest(okHttpRequestBuilder, request);
+
+        com.squareup.okhttp.Request okHttpRequest = okHttpRequestBuilder.build();
+        Call okHttpCall = client.newCall(okHttpRequest);
+        Response okHttpResponse = okHttpCall.execute();
+
+        StatusLine responseStatus = new BasicStatusLine(parseProtocol(okHttpResponse.protocol()), okHttpResponse.code(), okHttpResponse.message());
```
viniciusboson, line 91: This line should be indented. It has more than 100 characters.

app/src/main/java/com/circle/android/api/OkHttpStack.java
viniciusboson: This line is longer than 100 characters. it would be better if you indent it.

app/src/main/java/com/circle/android/api/OkHttpStack.java
```diff
@@ -0,0 +1,183 @@
...
+        entity.setContentLength(body.contentLength());
+        entity.setContentEncoding(r.header("Content-Encoding"));
+
+        if (body.contentType() != null) {
+            entity.setContentType(body.contentType().type());
+        }
+        return entity;
+    }
+
+    @SuppressWarnings("deprecation")
```
G1DR4: what exactly is deprecated?

app/src/main/java/com/circle/android/api/OkHttpStack.java
zestyping, line 121: You're right that this doesn't conform to our style guide.  However, this is open source code written by someone else, which is usually better to change as little as possible, in case the original code changes and we want to apply diffs.  (In fact, there were a couple of other small changes I made to the original, such as a spelling correction, but on further thought it's probably best to back them out, so I have undone my changes.)

llvasconcellos:
I want this!!!

zestyping:
This is ready for approval and merging.

PR #40, opened 2015-10-01
Added validations to New Patient Dialog

### LINKED ISSUE #31: Unexplained crash (bug report attached)
This happened; I don't know why.

Bug report: https://www.dropbox.com/s/osa648gwmei8be0/bugreport-2015-09-24-23-33-06.txt?dl=0

llvasconcellos:
Added validations to New Patient Dialog.
Renamed NewPatientActivityTest to EditPatientDialogFragmentTest.
This fixes #31

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
@@ -154,11 +154,9 @@ public class EditPatientDialogFragment extends DialogFragment {
             }
         }
 
-
         if (!valid) {
             throw new ValidationException("Validation Error");
         }
-
 
         Utils.logUserAction("patient_submitted",
             "id", id,
--- a/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
@@ -21,6 +21,7 @@ import android.support.v4.app.DialogFragment;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.WindowManager;
+import android.widget.Button;
 import android.widget.EditText;
 import android.widget.RadioButton;
 import android.widget.RadioGroup;
@@ -111,7 +112,7 @@ public class EditPatientDialogFragment extends DialogFragment {
         }
     }
 
-    public void onSubmit() {
+    public void onSubmit() throws ValidationException {
         String id = Utils.toNonemptyOrNull(mId.getText().toString().trim());
         String givenName = Utils.toNonemptyOrNull(mGivenName.getText().toString().trim());
         String familyName = Utils.toNonemptyOrNull(mFamilyName.getText().toString().trim());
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -174,18 +198,40 @@ public void onSubmit() {
...
         populateFields(args);
 
-        AlertDialog dialog = new AlertDialog.Builder(getActivity())
+        final AlertDialog dialog = new AlertDialog.Builder(getActivity())
             .setCancelable(false) // Disable auto-cancel.
             .setTitle(title)
-            .setPositiveButton(getResources().getString(R.string.ok), new DialogInterface.OnClickListener() {
+            .setPositiveButton(getResources().getString(R.string.ok), new DialogInterface
+                .OnClickListener() {
                 @Override public void onClick(DialogInterface dialogInterface, int i) {
```
G1DR4, line 204: is this override empty now?

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -119,23 +120,46 @@ public void onSubmit() {
...
+                case R.id.patient_sex_female:
+                    sex = Patient.GENDER_FEMALE;
+                    break;
+                case R.id.patient_sex_male:
+                    sex = Patient.GENDER_MALE;
+                    break;
+            }
         }
+
+
```
viniciusboson: Here should be only one empty line.

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -119,23 +120,46 @@ public void onSubmit() {
...
+        if (!valid) {
+            throw new ValidationException("Validation Error");
+        }
+
```
viniciusboson, line 160: Here has an extra empty line, as well.

zestyping:
Let's hold off on adding patient validation for now.  My reasons for not doing this are:
- We do want to allow patients to have no specified gender, and to let clinicians go back and fill in the gender later by editing the patient.
- By embedding the limit of 120 years into our code, we create coupling between the internals of our client and the internals of OpenMRS.  The 120-year limit is undocumented behaviour of the OpenMRS API, so we shouldn't write our code to assume it.  Instead, we should handle error messages returned by OpenMRS gracefully, giving good feedback to the user if it's a problem that the user can correct.
- Even if we do want to do validation, it's clunky to check on submit and then fail with an error message; it's better to do validation live as the user types, and disable the submission button if the input is invalid. 
- Perhaps more controversially, I think OpenMRS should not be limiting the age to 120 years at all.  I think that it was a poor design decision to embed this restriction into their API, that 120 years is an arbitrary choice, and that it is not the storage layer's job to enforce business logic restrictions like that.  OpenMRS should trust its clients — let the clients apply or not apply restrictions as best suits their purpose, and let the clients provide the UI necessary to make those restrictions make sense to users.

More importantly, at the moment, this isn't a high enough priority right now to spend time on it.  We don't need it for the field trip.

llvasconcellos:
@zestyping You are right!

PR #41, opened 2015-10-02
New patient validation tests

llvasconcellos:


zestyping:
As mentioned in the Slack discussion and in #40, let's hold back on patient validation for now.

PR #42, opened 2015-10-02
Fix/location list activity test

llvasconcellos:


app/src/androidTest/java/org/projectbuendia/client/ui/lists/LocationListActivityTest.java
```diff
@@ -55,31 +55,33 @@ public void testZonesAndTentsDisplayed_afterSettingsView() {
...
-        inUserLoginInitDemoPatient();
-        inUserLoginGoToLocationSelection();
-        inLocationSelectionClickLocation(LOCATION_NAME);
-        inPatientListClickFirstPatient(); // open patient chart
-
-        pressBack(); // back to search fragment
-        pressBack(); // back to tent selection screen
+        inUserLoginGoToPatientCreation();
+        screenshot("Test Start");
+        String id = generateId();
```
G1DR4, line 60: you don`t need id in any other place, generate it inside populateNewPatientFields()

app/src/androidTest/java/org/projectbuendia/client/ui/FunctionalTestCase.java
```diff
@@ -437,4 +441,37 @@ protected void inPatientListClickPatientWithId(String id) {
             .inAdapterView(hasId(R.id.fragment_patient_list))
             .atPosition(0));
     }
+
+    /** Generates IDs to identify the newly created patient. */
+    protected String generateId() {
+        String id = Long.toString(new Date().getTime()%100000);
+        return id;
```
zestyping: id is only used once; simplify to just `return <expr>`

`"" + x` is more concise than `Long.toString(x)`

zestyping:
The changes to make LocationListActivityTest pass look good. After separating these changes from the patient validation code (which shouldn't go in yet), feel free to merge.

zestyping:
Note, the reason not to commit the patient validation code is to save time.  If it will actually take less time to commit these as is without removing the patient validation code, it's okay to proceed with the more expedient option, with the understanding that the patient validation code will be backed out later.

app/src/androidTest/java/org/projectbuendia/client/ui/lists/LocationListActivityTest.java
llvasconcellos, line 60: Actually I need it in other tests.

app/src/androidTest/java/org/projectbuendia/client/ui/FunctionalTestCase.java
llvasconcellos: Done.

PR #52, opened 2015-10-06
Mark all strings needing internationalization

zestyping:


zestyping:
This is not a high priority — just publishing this as it may be useful.

PR #65, opened 2015-10-10
Fixing Chart Colors and Severity Bars.

### LINKED ISSUE #64: Severity Bars not showing on chart

llvasconcellos:
This fixes #64
Simple fix to a difficult to find bug.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/ObsFormat.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/ObsFormat.java
@@ -456,7 +456,7 @@ public class ObsFormat extends Format {
                 case "":
                 case "=":
                 case "==":
-                    return Objects.equal(value, operand);
+                    return Objects.equal(value.uuid, operand.uuid);
                 case "<":
                     return value.compareTo(operand) < 0;
                 case "<=":
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
@@ -192,6 +192,8 @@ public class PebbleExtension extends AbstractExtension {
                 for (Object item : (Collection) input) {
                     values.add(item instanceof ObsValue ? (ObsValue) item : null);
                 }
+            } else if (input instanceof ObsValue) {
+                values.add((ObsValue) input);
             }
             return formatValues(values, asFormat(args.get("format")));
         }


app/src/main/java/org/projectbuendia/client/ui/chart/ObsFormat.java
```diff
@@ -456,7 +456,7 @@ private boolean matches(@Nullable Object obj, String operator, String operandStr
                 case "":
                 case "=":
                 case "==":
-                    return Objects.equal(value, operand);
+                    return Objects.equal(value.uuid, operand.uuid);
```
zestyping, line 459: This doesn't look right to me.  If we let ObsValue.equals() do the comparison, it will check all its fields.  With this change, all numbers, strings, or dates will compare equal even though they shouldn't.

What was the root cause of the incorrect behaviour you were seeing, and how did this change fix it?

app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
```diff
@@ -192,6 +192,8 @@ public Object apply(Object input, Map<String, Object> args) {
                 for (Object item : (Collection) input) {
                     values.add(item instanceof ObsValue ? (ObsValue) item : null);
                 }
+            } else if (input instanceof ObsValue) {
```
zestyping, line 195: I'm curious, how did this change help fix the problem?

I ask because it was the original intention for this filter to not handle an individual value; format_value is supposed to take an individual value and format_values is supposed to take a list or array of values.

app/src/main/java/org/projectbuendia/client/ui/chart/ObsFormat.java
llvasconcellos, line 459: Yes,  probably this not a right fix for the problem.  The thing is that operand.name was  null.

app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
llvasconcellos, line 195: Sometimes input  is a instance of  ObsValue not an array  or collection.

app/src/main/java/org/projectbuendia/client/ui/chart/ObsFormat.java
zestyping, line 459: Oops, it looks like this line 459 change got committed.  Would you mind reverting it?

app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
zestyping, line 195: When is the input a single instance of ObsValue?  Can that call site use format_value instead of format_values?

PR #66, opened 2015-10-10
Fix order execution counts always appearing as zero.

zestyping:
This fixes a bug that caused the order execution counts in the grid cells to always be displayed as zero.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
@@ -158,11 +158,9 @@ public class ChartRenderer {
         void addObservations(List<Obs> observations) {
             for (Obs obs : observations) {
                 if (obs == null) continue;
-                ObsPoint point = obs.getObsPoint();
-                if (point == null) continue;
 
-                mDays.add(new DateTime(point.time).toLocalDate());
-                Column column = getColumnContainingTime(point.time);
+                mDays.add(new DateTime(obs.time).toLocalDate());
+                Column column = getColumnContainingTime(obs.time);
 
                 if (obs.conceptUuid.equals(AppModel.ORDER_EXECUTED_CONCEPT_UUID)) {
                     Integer count = column.executionCountsByOrderUuid.get(obs.value);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -158,11 +158,9 @@ public void render(Chart chart, Map<String, Obs> latestObservations,
         void addObservations(List<Obs> observations) {
             for (Obs obs : observations) {
                 if (obs == null) continue;
-                ObsPoint point = obs.getObsPoint();
-                if (point == null) continue;
```
viniciusboson, line 162: Ping, 
i didn't understand this change quiet well this change. Could you explain me better? 
I`ve noticed that this was point which was blocking the counting but  instead of removing it, i'm was trying to assign an ObsPoint to the executed order. Isn't  that make sense ?

viniciusboson:
This PR correct previous counted executed orders but the count doesn't work for the current day.
If you execute an order and then refresh the chat, the count remains zero

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
zestyping, line 162: The problem here is that getObsPoint returns null if there is no data type.  The "order executed" observations have no data type, so this is skipped and the order executions do not get counted.  But this works fine if we use obs.time directly instead of converting to ObsPoint first.

I moved the null check of the ObsPoint to addObs.  It's better anyway to protect near the site of the actual call that is sensitive to a null value, i.e. the call to add().

llvasconcellos:
@viniciusboson Actually that worked for me. I'm merging it. If you see this happening again please create a issue.

viniciusboson:
I've created a new patient (887766 ID) and then i`ve recorded the treatment, and the result was as expected.
screenshot_2015-10-14-18-54-21
screenshot_2015-10-14-18-54-31
screenshot_2015-10-14-18-54-59

But when i`ve tried to record a treatment for an existing patient, the count was not updating.
This happened with several patients.
screenshot_2015-10-15-18-59-20
screenshot_2015-10-15-18-59-45
screenshot_2015-10-15-19-00-10

PR #67, opened 2015-10-10
Fix order cells not being clickable for orders with no stop time.

zestyping:
This fixes a bug that caused no number to be displayed in grid cells for orders with no stop time (thus making it impossible to click to bring up the order execution dialog).

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -117,7 +117,7 @@
           {% for column in columns %}
             <td class="{{column.start == nowColumnStart ? 'now' : ''}}">
               {% set active = intervals_overlap(order.interval, column.interval) %}
-              {% if order.stop == null and previousActive %}
+              {% if future and order.stop == null %}
                 <div class="future">&nbsp;</div>
               {% elseif order.stop != null and previousActive and not active %}
                 <div class="stop">Stop</div>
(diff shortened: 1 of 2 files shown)


viniciusboson:
theses changes look fine to me

llvasconcellos:
After merging PR #66 I did not see this problem occur. What am I missing?

G1DR4:
checking...

llvasconcellos:
I'm closing since this fix isn't needed.

zestyping:
The change in ChartRenderer.java is not relevant after feature/ping-column-gaps is applied.

The change in chart.html is significant, though — the specific case it handles is when an order has no stop time, and the current day is after the start time of the order.

For example, suppose the order has start = Jan 1 and stop = null, and today's date is Jan 2.

Before applying this change, consider the loop over the columns:
- Jan 1: previousActive is false, active is true, future is false, order.stop is null — render the `class="past active"` branch (a clickable number)
- Jan 2: previousActive is true, active is true, future is false, order.stop is null — render the `class="future"` branch (a space)
- Jan 3: previousActive is true, active is true, future is true, order.stop is null — render the `class="future"` branch (a space)
  This is a problem because the user can't click on today's cell (Jan 2) to record that the order is given today.

After applying this change:
- Jan 1: previousActive is false, active is true, future is false, order.stop is null — render the `class="past active"` branch (a clickable number)
- Jan 2: previousActive is true, active is true, future is false, order.stop is null — render the `class="past active"` branch (a clickable number)
- Jan 3: previousActive is true, active is true, future is true, order.stop is null — render the `class="future"` branch (a space)
  So this should cause a clickable number to appear for today, which the user can increment, but no numbers to appear going into the future, because the order has no stop date.

PR #68, opened 2015-10-12
Remove redundant JS code in chart.html.

zestyping:
The popup() function is already defined in chart.js.

No need to use document.ready — just execute the JS at the end.

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -144,32 +144,6 @@
         </th>
       </tr>
     </tbody>
-    <script>
-      function popup(name, values) {
-        var dialog = document.getElementById('obs-dialog');
-        var html = '<h2>' + name + '</h2>';
-        var text = name + '\n\n';
-        if (values.length > 0) {
-          for (var i = 0; i < values.length; i++) {
-            var value = values[i][0], observed = values[i][1];
-            html += '<b>' + value + '</b> at ' + observed + '<br>';
-            text += value + ' at ' + observed + '\n';
-          }
-          html += '&nbsp;<br>Comments <input type="text" size="40">';
-        } else {
-          html += 'No observations.';
-          text += 'No observations.';
-        }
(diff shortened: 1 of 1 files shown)


viniciusboson:
this change looks fine to me

G1DR4:
ok

app/src/main/assets/chart.html
```diff
@@ -190,9 +164,8 @@
     {% endif %}
   {% endfor %}
 
-  $(document).ready(function() {
-    $('#grid').freezeHeader({top: true, left: true});
-  });
+  $('#grid').freezeHeader({top: true, left: true});
```
llvasconcellos: When using jquery to alter DOM is good practice to wrap code inside the document ready event handler like this:

``` javascript
$(document).ready(function(){ 
    $("table").freezeHeader(); 
});
```

This also prevents erros. The problem is that although the script tag is in the bottom of the page, this does not mean that the html is rendered before the script execution. 
So, a good overall practice is always use _$(document).ready(function(){});_ for all Javascript code to be executed after page load.

Do this at least for the jQuery related code.

llvasconcellos: One other thing: I know that this does not concern this pull request in particular but the line bellow:

``` javascript
var data = {{dataCellsByConceptId | raw}};
```

Is generating a `Uncaught SyntaxError: Unexpected token ;`. This is because it's being rendered like this:

``` javascript
var data = ;
```

I noticed that dataCellsByConceptId is not being used anywhere else in the project so it will always be empty.
Reading runTileScript and runChartRowScript functions I noticed that both expects _data_ to be an array so a good fix would also be to use hard coded brackets to avoid syntax errors. Like this:

``` javascript
var data = [{{dataCellsByConceptId | raw}}];
```

llvasconcellos: OK. I saw PR #71!

PR #69, opened 2015-10-12
Fix order sync

zestyping:
Orders were not getting automatically synced at all, previously.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/sync/SyncAccountService.java
+++ b/app/src/main/java/org/projectbuendia/client/sync/SyncAccountService.java
@@ -66,12 +66,8 @@ public class SyncAccountService extends Service {
             ContentResolver.setIsSyncable(account, Contracts.CONTENT_AUTHORITY, 1);
             ContentResolver.setSyncAutomatically(account, Contracts.CONTENT_AUTHORITY, true);
             Bundle b = new Bundle();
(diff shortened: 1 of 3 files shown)


viniciusboson:
theses changes look fine to me

G1DR4:
ok to me too

llvasconcellos:
I don't see any difference after merging this fix to dev and it seems that it still missing something.

I clicked in "ADD A NEW TREATMENT" and filled the form for a new medicine. After I saved a new encounter is created in the patient's visits tab on OpenMRS. But the location, provider and observation is missing as you can see bellow:

image

But when I add a new _order executed_ (mark as given now) then on OpenMRS is created another encounter but this time the location and observation is informed only missing the provider name. See bellow:

image

**Again, I don't see any difference with or without this fix.**

llvasconcellos:
During my tests, at one point, all the "order executed" became zeroes again but in OpenMRS they were still there. You can see this on patient ID number 5.

llvasconcellos:
@capnfabs Can you take a look at this?

capnfabs:
The code looks 👌 to me; I think the issues you're observing [above](#issuecomment-148270559) are unrelated to this change specifically.

Can you clarify what the issue is @llvasconcellos? I'd be inclined to merge this (it's an improvement) and then address those issues separately.

PR #70, opened 2015-10-12
Fix bug in storing order stop times on the server.

zestyping:


viniciusboson:
this change looks fine to me

G1DR4:
ok to me too

llvasconcellos:
Works perfectly.

PR #71, opened 2015-10-12
Fix a bad merge that omitted the setting of dataCellsByConceptId.

zestyping:


viniciusboson:
this change seems ok to me

G1DR4:
ok to me too

llvasconcellos:
Ooops. A bad merge found. I'm afraid there might be more...

PR #72, opened 2015-10-12
Fix insertion of empty columns; display column gaps.

zestyping:
This determines the columns to drop a bit more efficiently, and also displays a grey gap to show that columns are missing.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
@@ -11,7 +11,6 @@ import com.mitchellbosecke.pebble.PebbleEngine;
 import org.joda.time.Chronology;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
-import org.joda.time.Instant;
 import org.joda.time.LocalDate;
 import org.joda.time.ReadableInstant;
 import org.joda.time.chrono.ISOChronology;
@@ -32,7 +31,6 @@ import org.projectbuendia.client.utils.Utils;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -102,7 +100,8 @@ public class ChartRenderer {
         List<String> mTileConceptUuids;
         List<String> mGridConceptUuids;
         List<Order> mOrders;
-        LocalDate mToday;
+        DateTime mNow;
+        Column mNowColumn;
         LocalDate mAdmissionDate;
         LocalDate mFirstSymptomsDate;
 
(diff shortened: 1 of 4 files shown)


viniciusboson:
These changes seem ok to me

G1DR4:
ok, just don`t forget regional date settings and i18n too

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -153,15 +152,12 @@ public void render(Chart chart, Map<String, Obs> latestObservations,
...
         }
 
         void addObservations(List<Obs> observations) {
             for (Obs obs : observations) {
                 if (obs == null) continue;
-                ObsPoint point = obs.getObsPoint();
-                if (point == null) continue;
-
-                mDays.add(new DateTime(point.time).toLocalDate());
                 Column column = getColumnContainingTime(point.time);
```
llvasconcellos: @zestyping This merge is conflicting right here. And if I remove this lines point is not declared.
Can you please resolve this conflict?

llvasconcellos:
Waiting for @zestyping to have a look.

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
zestyping: Fixed now and repushed (which updated this PR).  After removing these four lines, if you replace `point.time` with `obs.time` it works out.

zestyping:
Oops, should have commented over here.  I fixed the conflict and repushed, so this PR is able to be merged now.

llvasconcellos:
Sorry @zestyping I really want to merge this but I noticed something. This Pull Request ignores orders with specific duration as you can see on the screenshots bellow:

_BEFORE:_
device-2015-10-29-102951

_AFTER:_
device-2015-10-29-102632

llvasconcellos:
One other thing I noticed is that day 1, 2 and 3 are totally empty. Shouldn't they be collapsed?

dancunningham:
@llvasconcellos maybe in that example day 1 only appears empty but it actually has an observation for "Date of admission of Ebola treatment centre" right?

I think we can accept that edge case; it's unlikely a patient will be created and no obs entered for 3 days.

llvasconcellos:
@dancunningham How about the missing day 8 and 9 that should be triggered by a doctor order with end date? I think this is a bug. If I merge this right now we will break this functionality.

zestyping:
You're correct that it's a bug — will look at this later today.

zestyping:
And Dan's explanation of the blank columns sounds right to me.

zestyping:
Okay, I see what happened.  The method addOrders() was left out of this
pull request.  You can see the code for addOrders(), and the call to it, in
the scratch/chad-bokoro-new-api branch.  I don't have access to my laptop
to revise the pull request at the moment; would you mind copying over that
method to here and merging?

llvasconcellos:
OK. We are almost there now, but one last thing is missing: The "stop" future day for the order. See the screen shots:

_THIS BRANCH:_
device-2015-11-04-083615

_DEV:_
device-2015-11-04-083902

@zestyping do you know what is missing?

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -168,7 +167,18 @@ void addObservations(List<Obs> observations) {
             }
         }
 
-        /** Returns a column for a given date. Creates a new column, if it does not exist yet */
+        /** Ensures that columns are shown for any days in which an order is prescribed. */
+        void addOrders(List<Order> orders) {
+            for (Order order : orders) {
+                if (order.stop != null) {
+                    for (DateTime dt = order.start; !dt.isAfter(order.stop.plusDays(1)); dt = dt.plusDays(1)) {
```
llvasconcellos, line 174: @zestyping I added the order.stop.plusDays(1) to make the STOP day appear for a future order. Is that ok? If so it will be ready to merge!

zestyping, line 174: Yes, that sounds right!  (I think the UI for this is not that great and we
might want to get rid of the STOP to make it more like the paper charts
they use.  But that's for later.)

PR #73, opened 2015-10-12
Use real minus signs when showing negative numbers.

zestyping:


zestyping:
This is more typographically proper—and also more visible than a hyphen.

viniciusboson:
this change seems ok to me

G1DR4:
is it colored in red?

PR #74, opened 2015-10-12
Remove the "Chart last updated" line.

zestyping:


zestyping:
This line is not a good use of space on the chart, and we decided to get rid of it for now.  We may find a different way to show this information later if it is really needed.

viniciusboson:
Theses changes look ok to me

G1DR4:
ok to me too

PR #76, opened 2015-10-12
Change ID button label to "Search by ID".

zestyping:
"ID" was too cryptic for users.

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -45,7 +45,7 @@
   <string name="family_name">Nom de famille</string>
   <string name="add_user_example_family_name">Smith</string>
   <string name="action_search">Rechercher</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values-fr/strings.xml
```diff
@@ -45,7 +45,7 @@
   <string name="family_name">Nom de famille</string>
   <string name="add_user_example_family_name">Smith</string>
   <string name="action_search">Rechercher</string>
-  <string name="action_go_to">ID</string>
+  <string name="action_go_to">Recherche ID</string>
```
viniciusboson, line 48: Not sure, but maybe 'Rechercher par ID' would be better. It worths double check it.

zestyping, line 48: This text was Ivan's suggestion, so I went with it.

G1DR4:
ok

PR #77, opened 2015-10-12
Fix size of frozen corner cell and mark frozen headers with CSS classes.

zestyping:


zestyping:
The symptom we were seeing here is that when you scroll up so that the top headers freeze, the top-left corner cell becomes too tall and covers up a bit too much of the leftmost column.

viniciusboson:
Theses changes look fine to me.

G1DR4:
it would be best if we fix this overflow-hidden table so that it works in different resolutions

PR #78, opened 2015-10-12
Internationalize age; show age in months only up to 59 months.

zestyping:
We got a specific request from the MedCo here to show ages in months (18 months instead of 1 year 6 months).

### DIFF
--- a/app/src/main/res/layout/listview_tent_header.xml
+++ b/app/src/main/res/layout/listview_tent_header.xml
@@ -24,6 +24,6 @@
         android:layout_height="wrap_content"
         android:text="@string/unknown_tent"
         android:textAppearance="@style/text.heading"
-        tools:text="@string/probable" />
+        tools:text="Probable 1 (10 patients)" />
 
 </LinearLayout>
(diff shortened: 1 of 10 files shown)


app/src/main/res/layout/listview_tent_header.xml
```diff
@@ -24,6 +24,6 @@
         android:layout_height="wrap_content"
         android:text="@string/unknown_tent"
         android:textAppearance="@style/text.heading"
-        tools:text="@string/probable" />
+        tools:text="Probable 1 (10 patients)" />
```
viniciusboson, line 27: Shouldn't it be internationalized ?

zestyping, line 27: The tools:text string is never shown to users -- it only appears as sample
text in the graphical UI layout editor in Android Studio.

G1DR4, line 27: ping, i`m fixing all the TODO i18n and also found even more strings to translate.

G1DR4, line 27: theres a feature/i18n branch

PR #80, opened 2015-10-16
Fix EditPatientDialogFragmentTest

llvasconcellos:
Removing chart last updated test according to PR #74


viniciusboson:
This change looks fine to me.

PR #81, opened 2015-10-16
Remove code from FunctionalTestCase

llvasconcellos:
Removed two methods after they were replaced by a better updated one.


viniciusboson:
theses changes look ok to me

PR #83, opened 2015-10-17
Updating FilteredPatientListActivityTest to the new UI.

llvasconcellos:


app/src/androidTest/java/org/projectbuendia/client/ui/lists/FilteredPatientListActivityTest.java
```diff
@@ -26,30 +26,30 @@ public void setUp() throws Exception {
...
-        waitForProgressFragment(); // Wait for patients.
-    }
-
-    /** Looks for two zone headings and at least one patient. */
+    /** Looks for one zone heading and at least one patient. */
     public void testZoneAndPatientDisplayed() {
         openPatientList();
         screenshot("Test Start");
-        // There should be patients in both Triage and S1.
-        expectVisibleSoon(viewThat(hasTextMatchingRegex("Triage \\((No|[0-9]+) patients?\\)")));
```
llvasconcellos, line 51: I couldn't make this work tried some unicode combinations and non break spaces.
I then realized that this tested for a number of patients or no patients at all, so why not just test to see if the heading is there using hastTextContaining? That's what I did then.

app/src/androidTest/java/org/projectbuendia/client/ui/lists/FilteredPatientListActivityTest.java
```diff
@@ -26,30 +26,30 @@ public void setUp() throws Exception {
...
     }
 
+    /** Opens the patient list. */
+    private void openPatientList() {
+        waitForProgressFragment(); // Wait for tents.
+        click(viewWithId(R.id.location_selection_all_patients));
+        waitForProgressFragment(); // Wait for patients.
+    }
+
     /** Looks for the filter menu. */
```
viniciusboson, line 36: private methods should be after the public ones.

Issue #85, opened 2015-10-17
Temperature is not being updated on patient chart.


llvasconcellos:
This was caught by PatientChartActivityTest -> testEncounter_latestEncounterIsAlwaysShown().

When you record a temperature more two or more times in a row some of it's values goes missing.

dancunningham:
Tested with 3 updates to temperature within a minute and it seems to work fine, even after doing a pull down refresh from patient list. Closing this!

capnfabs:
@dancunningham I reckon you're still observing the same problem in @viniciusboson's PR #172. But we're also tracking this in #125 so let's keep tracking it there.

PR #86, opened 2015-10-17
Added a wait timeout to give time to chart reload.

llvasconcellos:
Note that testEncounter_latestEncounterIsAlwaysShown does not pass because of issue #85.
I tried to implement IdlingResource on WebView with partial success. I created a branch for it and will continue in a more suitable time.


viniciusboson:
@llvasconcellos, does this change solve that intermittent that was occurring (ignoring the issue #85) ?
If not, maybe should be better to wait for a permanent solution (IdlingResource).

llvasconcellos:
It fixes the intermittent issue. It's just not the "perfect fix". I'm going with done is better then perfect.
"Feito é melhor do que perfeito" in portuguese.
Anyways, I'm going to get back to it later on.

PR #87, opened 2015-10-17
Add title tag to chart.html

llvasconcellos:
Added title tag to chart.html to a proper formed html and assist on Espresso's idlingResource webview implementation.


viniciusboson:
This change looks ok to me

PR #90, opened 2015-10-18
Fix incremental fetch missing observations

zestyping:
This is a pretty subtle bug due to OpenMRS yet again failing to store and retrieve data faithfully.  OpenMRS uses the Java Date object for representing timestamps, but it does not preserve Date objects correctly—it truncates the fractional seconds in storage, leading to bugs like this one.

There is a corresponding fix on the server side, but because the loss of observations is such a bad, user-visible bug, so it's worth doing a quick fix on the client side until the fix on the server is in production.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/net/OpenMrsChartServer.java
+++ b/app/src/main/java/org/projectbuendia/client/net/OpenMrsChartServer.java
@@ -100,8 +100,19 @@ public class OpenMrsChartServer {
         Instant lastTime,
         Response.Listener<JsonPatientRecordResponse> successListener,
         Response.ErrorListener errorListener) {
+
+        // TODO/cleanup: Remove this fix from the client once the equivalent fix in
+        // EncounterResource.filterEncountersByModificationTime is deployed in the server.
+
+        // Even though OpenMRS getDateCreated() and getDateModified() return Date objects that
+        // have millisecond precision, it truncates away the fractional milliseconds from these
+        // values when storing them in the database!!  Bad OpenMRS.  So for example, an observation
+        // created at 12:00:00.750 will be stored with a creation time of 12:00:00 -- which means
+        // that in order to return all the observations that were created after 12:00:00.500, we
+        // actually have to check for creation times after 12:00:00.000.
+        long startMillis = ((int) (lastTime.getMillis() / 1000)) * 1000;
         doEncountersRequest(mConnectionDetails.getBuendiaApiUrl()
-                + "/encounters?sm=" + lastTime.getMillis(),
+                + "/encounters?sm=" + startMillis,
             successListener, errorListener);
     }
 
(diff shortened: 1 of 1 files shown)


llvasconcellos:
I started reviewing and testing this but the client in my tablet is very unstable. With this branch the problem described in issue #85 still happens. I will pause my work on merging new PRs until I have the tablet working with my local server.

zestyping:
Strange.  This fix is what made behaviour like #85 go away for me -- it
seems to be more reliable at storing observations now.  Let me know if you
discover any clues!

llvasconcellos:
So, after I fixed the server on my local machine I have a better environment to test and yes, #85 still happens. If you enter observations one after another some goes missing. On the OpenMRS side, the observations are there but the order of events is not right.
After noticing that I cleared the data and did a full resync. Some of the Observations appeared but the 37 Celsius temperature was missing. The ordering is not right again as you can see in the screenshots bellow:

image

device-2015-10-26-164819

39 degrees should be the last one and 37 is missing.

llvasconcellos:
Just to add: I inserted this observations manually, maybe on the same minute but not on the same second.

zestyping:
Okay, interesting!  What do you mean by "manually" in this case?

llvasconcellos:
@zestyping I mean not by testing. There is a test that does observations one after the other. Which is failing by the way. (PatientChartActivityTest->testEncounter_latestEncounterIsAlwaysShown())

capnfabs:
I'm observing this too when running tests on the incremental sync branch (which hasn't explicitly addressed this issue yet).

The client makes requests: `http://127.0.0.1:9000/openmrs/ws/rest/v1/projectbuendia/encounters?since=2015-11-06T03:26:56.970Z`
and receives response

``` json
{"results":[],"snapshotTime":"2015-11-06T03:27:02.108Z"}
```

Manually requesting: `http://127.0.0.1:9000/openmrs/ws/rest/v1/projectbuendia/encounters?since=2015-11-06T03:26:56.000Z` (truncated to nearest whole second) results in

``` json
{
  "results": [
    {
      "uuid": "1c0f3be5-5aaf-4878-a2df-293ba8fecd9a",
      "patient_uuid": "5a403158-dc17-4a49-bad5-4e00123a6301",
      "timestamp": "2015-11-06T03:26:56.000Z",
      "observations": {
        "162622AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA": "2015-11-06"
      }
    }
  ],
  "snapshotTime": "2015-11-06T03:28:44.157Z"
}
```

llvasconcellos:
Nice catch Fabian!

capnfabs:
I've got a fix for this on the feature/incremental-sync branch; let's leave it until we merge that to avoid additional churn. ETA later this week. Feel free to reopen if you think it's worth fixing sooner :)

Also, thanks for spotting the bug @zestyping, great find!

PR #91, opened 2015-10-18
Returning zero users is not an error condition.

zestyping:
You can always create a user if you need one.


viniciusboson:
This change looks OK to me.

llvasconcellos:
If no user is available a new one is created. But ok.

zestyping:
Yeah it's basically unused code.  Always feels good to clean up code we
don't need. :)

PR #92, opened 2015-10-18
Add an ID prefix field for patients.

zestyping:
We added this field because of the way that facility codes become part of patient IDs in the Bokoro project.  In this project IDs are quite long due to Ministry of Health conventions, like HAD/BOK/BOK/123, even though the first part (HAD/BOK/BOK) is highly redundant and something like BOK/123 alone would suffice in practice.

This implementation assumes that IDs will look like XYZ/123 where XYZ is a facility prefix and 123 is the numeric patient ID.  We haven't validated that this will actually work for users in Bokoro yet, but we know that numbers alone are not sufficient and we needed an interim solution to disambiguate patients from different facilities.  I think it doesn't hurt to review this anyway, and possibly merge it this week, even though we will probably change our approach to this later.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
@@ -62,7 +62,7 @@ public class EditPatientDialogFragment extends DialogFragment {
     @InjectView(R.id.patient_sex_female) RadioButton mSexFemale;
     @InjectView(R.id.patient_sex_male) RadioButton mSexMale;
 
-    private static final Pattern ID_PATTERN = Pattern.compile("([a-zA-Z]+)/?([0-9]+)");
+    private static final Pattern ID_PATTERN = Pattern.compile("([a-zA-Z]+)/?([0-9]+)*");
 
     private LayoutInflater mInflater;
     @Nullable private ActivityUi mActivityUi;  // optional UI for showing a spinner
--- a/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
@@ -14,7 +14,9 @@ package org.projectbuendia.client.ui.dialogs;
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.content.DialogInterface;
+import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.app.DialogFragment;
@@ -37,6 +39,9 @@ import org.projectbuendia.client.models.Patient;
 import org.projectbuendia.client.models.PatientDelta;
 import org.projectbuendia.client.utils.Utils;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import javax.inject.Inject;
 
 import butterknife.ButterKnife;
@@ -47,6 +52,7 @@ public class EditPatientDialogFragment extends DialogFragment {
     @Inject AppModel mModel;
     @Inject CrudEventBus mCrudEventBus;
 
+    @InjectView(R.id.patient_id_prefix) EditText mIdPrefix;
     @InjectView(R.id.patient_id) EditText mId;
     @InjectView(R.id.patient_given_name) EditText mGivenName;
     @InjectView(R.id.patient_family_name) EditText mFamilyName;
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/listview_cell_search_results.xml
```diff
@@ -33,14 +33,14 @@
...
-        android:width="108sp"
+        android:width="120sp"
         android:background="@android:color/darker_gray"
         android:gravity="center"
         android:paddingBottom="12sp"
         android:paddingTop="12sp"
         android:textAppearance="@style/text.large"
         android:textColor="@color/white"
-        tools:text="Kh.1234"/>
+        tools:text="BOK/1234"/>
```
viniciusboson, line 43: Does it need to be hardcoded ?

zestyping, line 43: This string is just for us to see when we look at a preview rendering of the UI in Android Studio.  We just need to provide some sort of example string there, so that the UI elements have approximately the right size to lay them out.

llvasconcellos:
I like the idea of the prefix persisting to the next new patient but I still feel that some kind of validation would be good. Not for validation but for feed back on the required fields.
For instance, I can't add new patients without prefix and this should be visible to the user.

llvasconcellos:
I tested and the first time I could create an patient only giving the ID prefix but when editing this patient the prefix goes to the id field because there where no ID.

llvasconcellos:
I think It's a good implementation and even if we need to change you already did almost everything. 
We need to solve some minor issues and update the functional test for it.
I'm coming back for this one later.

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -56,6 +62,8 @@
     @InjectView(R.id.patient_sex_female) RadioButton mSexFemale;
     @InjectView(R.id.patient_sex_male) RadioButton mSexMale;
 
+    private static final Pattern ID_PATTERN = Pattern.compile("([a-zA-Z]+)/?([0-9]+)");
```
capnfabs: Seeing as you're relying upon the regex not to match for ids without a prefix, you might want to just go with `"([a-zA-Z]+)/([0-9]+)"` for this regex (i.e. make the `/` required).

llvasconcellos:
@dancunningham What's your input on this?

dancunningham:
I noticed a bug on this - when you go to edit a patient the prefix ends up in the ID field, but isn't editable as that's a numeric-only field in the UI.

llvasconcellos:
@dancunningham That's what I mentioned on https://github.com/projectbuendia/client/pull/92#issuecomment-151333327
I'm fixing it right now.

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
llvasconcellos: Sorry @capnfabs, can you explain the benefit of removing the "?" ?

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -56,6 +62,8 @@
     @InjectView(R.id.patient_sex_female) RadioButton mSexFemale;
     @InjectView(R.id.patient_sex_male) RadioButton mSexMale;
 
+    private static final Pattern ID_PATTERN = Pattern.compile("([a-zA-Z]+)/?([0-9]+)*");
```
llvasconcellos, line 65: I added the \* in the end to match blank ids. This fixes the problem when editing the patient.
@dancunningham This fix the previous bug.

dancunningham:
Great!

So is this ready to merge?

llvasconcellos:
Yes.

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
zestyping, line 65: Did you intend to replace the `+` with `*` instead of tacking it on at the end?

`([0-9]+)*` seems strange, and it looks like one of those regular expressions that can blow up the regex engine (runtime exponential in the length of the input, e.g.
 it will backtrack over the string `123456789123456789123456789x` trying something like 30 million different paths).

capnfabs, line 65: It could also result in IndexOutOfBoundsExceptions, because it makes the second subcapture optional, and the code assumes that the second subcapture is available.

As Ping states, I think you're after `([a-zA-Z]+)/?([0-9]*)` instead.

PR #94, opened 2015-10-23
Changing time formatter to use specif time zone rather than UTC

viniciusboson:
The timestamp showed in patient's dialog was inaccurate. 
It was showing the timestamp in UTC rather than the user time zone.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
@@ -263,7 +263,9 @@ public class PebbleExtension extends AbstractExtension {
             String pattern = "" + args.get("pattern");
 
             if (input instanceof ReadableInstant) {
-                return DateTimeFormat.forPattern(pattern).print(new DateTime(input));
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/projectbuendia/client/ui/chart/PebbleExtension.java
```diff
@@ -263,7 +263,9 @@ static String formatValues(List<ObsValue> values, Format format) {
             String pattern = "" + args.get("pattern");
 
             if (input instanceof ReadableInstant) {
-                return DateTimeFormat.forPattern(pattern).print(new DateTime(input));
+                //Return time specif to time zone rather than UTC
+                return DateTimeFormat.forPattern(pattern).print(
```
zestyping, line 267: Yes, this works, and it's even nicer than what I suggested.  To make it clearer why we're doing this, though, specify the time zone like so:

```
new DateTime(input, DateTimeZone.getDefault())
```

This makes it obvious that we want to format the timestamp in the default time zone.

zestyping:
Please name your branch in the form `feature/viniciusboson-`_description_ .

PR #95, opened 2015-10-23
Fixing time formatter to use specific time zone rather than UTC

viniciusboson:
Renamed branch, related to PR #94.

I've also added a unit test for this change.


llvasconcellos:
Such a elegant fix. Well done @viniciusboson!

PR #96, opened 2015-10-26
Fix a ClassCastException when trying to sync patients.

capnfabs:
I'm not sure if this is only happening for my `javac` (currently OSX-supplied) but I'm getting `ClassCastException`s owing to `javac` not being able to infer generic types properly. This PR fixes that.

If you've got any other ideas (e.g. installing a different `java`) please let me know; the old code doesn't seem like it should cause crashes to me.

`java -version`:

```
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/models/AppModel.java
+++ b/app/src/main/java/org/projectbuendia/client/models/AppModel.java
@@ -111,9 +111,10 @@ public class AppModel {
     public void fetchPatients(CrudEventBus bus, SimpleSelectionFilter filter, String constraint) {
         bus.registerCleanupSubscriber(new CrudEventBusCleanupSubscriber(bus));
         // NOTE: We need to keep the object creation separate from calling #execute() here, because
-        // the type inference breaks otherwise, which throws
+        // the type inference breaks on Java 8 otherwise, which throws
         // `java.lang.ClassCastException: java.lang.Object[] cannot be cast to java.lang.Void[]`.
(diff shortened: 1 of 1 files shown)


viniciusboson:
Hi @capnfabs,

You are right. This always happens when compiled with java 8 version.
Nice fix.

llvasconcellos:
I'm confused. I thought that Java 7 was mandatory. 
On the ReadMe we have prerequisite JDK 7, but reading more carefully I saw that it says "does not report a version >= 1.7". This indicates that maybe we could use 8 but if so we need to review the documentation.

capnfabs:
I was super confused by that statement as well. I think it says "install JDK 7 if you don't have version 7 or newer".

So I don't think it's at odds with this change. But we should clarify the language there, and I will do so. I think it's good to ensure that we are compatible with compilation by Java 8, just to make initial setup easier for developers.

llvasconcellos:
I agree. On the server there is an issue with Java 8 as well.

capnfabs:
Yes, that's correct. If I recall correctly, it's a _runtime_ issue instead of a _compile time_ issue, and it's a problem in OpenMRS. I think I had to set JAVA_HOME to the Java 7 runtime when trying to run the OpenMRS server.

llvasconcellos:
OK. So I built and ran using Java 7 with no problems. I'm ok to merge this PR.

PR #97, opened 2015-10-27
Update README:

capnfabs:
- Clarify wording around JDK7 vs JDK8, see discussion on
  https://github.com/projectbuendia/client/pull/96
- Change `git clone` command to use the `ssh` protocol instead of `https`. This
  makes pushing commits easier later, because the SSH remote can use device
  certs instead of Github asking for your password on push.


llvasconcellos:
Fine by me.

PR #99, opened 2015-10-27
Dead code removal

capnfabs:


app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -98,16 +87,16 @@ private void populateFields(Bundle args) {
         LocalDate birthdate = Utils.toLocalDate(args.getString("birthdate"));
         if (birthdate != null) {
             Period age = new Period(birthdate, LocalDate.now());
-            mAgeYears.setText("" + age.getYears());
-            mAgeMonths.setText("" + age.getMonths());
+            mAgeYears.setText(String.valueOf(age.getYears()));
```
zestyping, line 90: Why do this?  Isn't `String.valueOf(x)` just a more verbose way to write `"" + x`?

capnfabs, line 90: It results in the same output, but it's semantically cleaner.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -420,30 +386,6 @@ private String getFormattedPcrString(double pcrValue) {
         }
 
         @Override public void updatePatientConditionUi(String generalConditionUuid) {
```
zestyping, line 388: Quick note: all the code related to (a) selecting the general condition with a specialized dialog and (b) tapping a tile to navigate directly to that section of the form was commented out but left in the code because it is functionality that the Ebola deployment depended on.  When we switched to WebView rendering, this code no longer compiled, as it was written for the original Android widget layouts.  It is commented out to show that although it is currently never called, the removal is temporary in the sense that without this code, the Ebola case suffers a regression (while worth the gains from using a WebView) by losing functionality for which we never made a product decision that it was okay to remove.

capnfabs, line 388: No worries. I'm happy to rollback this change if you don't think it's worthwhile; I was just cleaning stuff up whilst looking at the code.

llvasconcellos:
@capnfabs Can you leave the commented Ebola code and just PR the other changes?

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -355,7 +321,7 @@ private String getFormattedPcrString(double pcrValue) {
                     .color(0x00000000)
                     .sizeDp(36));
             if ((pcrLObservation == null || pcrLObservation.valueName == null)
-                && (pcrNpObservation == null || pcrNpObservation == null)) {
+                && (pcrNpObservation == null || pcrNpObservation.valueName == null)) {
```
llvasconcellos, line 324: What's happening here?

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
```diff
@@ -98,16 +87,16 @@ private void populateFields(Bundle args) {
...
+            mAgeMonths.setText(String.valueOf(age.getMonths()));
         }
-        Integer gender = args.getInt("gender");
-        if (gender != null) {
-            if (gender == Patient.GENDER_FEMALE) {
+        switch (args.getInt("gender", Patient.GENDER_UNKNOWN)) {
```
llvasconcellos, line 93: Why switch instead of if?

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
zestyping, line 90: Just a style difference then.  :)  Given two equivalent ways to express the same thing I'll usually pick the shorter one (less probability of causing a line break, which has a high cost).

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
zestyping, line 388: Not necessarily bad -- after all the code is always retrievable from git.  Just didn't want the knowledge to disappear with the code.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
capnfabs, line 324: I'm not sure what was going on previously, but it was definitely a typo that triggered an inspection in my IDE, which is why I looked into it. The logic for pcrNpObservation is supposed to mirror pcrLObservation here, but .valueName had been left off erroneously.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
capnfabs, line 388: Acknowledged. I'm inclined to track this with an issue instead of just by commenting out the relevant code, given that reimplementation would be completely different in the future. Thanks for the information!

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
capnfabs, line 90: Yeah, just a style difference. I think an inspection in AS had highlighted it, which is why I changed it.

app/src/main/java/org/projectbuendia/client/ui/dialogs/EditPatientDialogFragment.java
capnfabs, line 93: The idea here was:
- `args.getInt("gender")` doesn't return an `Integer`, it returns an `int`, and returns the value `0` if the bundle didn't contain the specified key.
- The old code only worked incidentally because `Patient.GENDER_UNKNOWN == 0`. 
- I changed it to a `switch` instead of an `if` because I thought it was more readable - `switch` implies that you're handling all possible cases, whereas `if` / `else if` is more readable for isolated checks. i.e. if we were just checking against one of the three possible values, I would have left it as an if, but seeing as we're checking against all possible values, it made more sense as a switch.

capnfabs:
@llvasconcellos I filed issue #111 to track the Ebola regression, this PR is ok as-is now.

capnfabs:
Ready to merge. :+1:

llvasconcellos:
Fine by me!

Issue #103, opened 2015-10-28
Client Crash when no Profile Applied.


llvasconcellos:
This happens right after you setup a new development server. 
As there is no profile applied to the new server some of the information is missing and the client crashes.

PR #105, opened 2015-10-28
Fix crash when no profile applied

llvasconcellos:
The app crashes as described in issue #103. 
Now the client does not crash but displays an empty chart with and no forms. 
This way the user can see clearly that there is something wrong with the profile or it's missing. 

device-2015-10-28-215316

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/sync/ChartDataHelper.java
+++ b/app/src/main/java/org/projectbuendia/client/sync/ChartDataHelper.java
@@ -211,37 +211,37 @@ public class ChartDataHelper {
                 String label = Utils.getString(c, ChartItems.LABEL, "");
                 if (parentRowid == null) {
                     // Add a section.
-
-
                     String SectionType = Utils.getString(c, ChartItems.SECTION_TYPE);
-
-
-                    switch (SectionType) {
-                        case "TILE_ROW":
-                            ChartSection tileGroup = new ChartSection(label);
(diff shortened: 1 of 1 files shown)


zestyping:
This is great!  Thanks for fixing this.

app/src/main/java/org/projectbuendia/client/sync/ChartDataHelper.java
```diff
@@ -211,32 +211,37 @@ public Chart getChart(String uuid) {
...
-                            tileGroups.add(tileGroup);
-                            tileGroupsById.put(rowid, tileGroup);
-                            break;
-                        case "GRID_SECTION":
-                            ChartSection rowGroup = new ChartSection(label);
-                            rowGroups.add(rowGroup);
-                            rowGroupsById.put(rowid, rowGroup);
-                            break;
+                    String SectionType = Utils.getString(c, ChartItems.SECTION_TYPE);
+                    if (SectionType != null) {
```
zestyping, line 215: Use lowercase for variable names like "sectionType"

capnfabs:
:+1: Nice work! Thanks for fixing this.

app/src/main/java/org/projectbuendia/client/sync/ChartDataHelper.java
llvasconcellos, line 215: You are right. Sometimes I wonder where was my mind when I do somethings like this...

Issue #107, opened 2015-10-29
Exception When clearing local data.


llvasconcellos:
![image](https://cloud.githubusercontent.com/assets/1710823/10823956/4118458e-7e45-11e5-83f8-09cabe9be195.png)

``` java
10-29 13:58:48.069  11829-11846/org.projectbuendia.client.dev E/Database﹕ close() was never explicitly called on database '/data/data/org.projectbuendia.client.dev/databases/buendia.db'
    net.sqlcipher.database.DatabaseObjectNotClosedException: Application did not close the cursor or database object that was opened here
            at net.sqlcipher.database.SQLiteDatabase.<init>(SQLiteDatabase.java:1945)
            at net.sqlcipher.database.SQLiteDatabase.openDatabase(SQLiteDatabase.java:881)
            at net.sqlcipher.database.SQLiteDatabase.openOrCreateDatabase(SQLiteDatabase.java:913)
            at net.sqlcipher.database.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:132)
            at net.sqlcipher.database.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:99)
            at org.projectbuendia.client.sync.Database.deleteDatabaseIfPasswordIncorrect(Database.java:190)
            at org.projectbuendia.client.sync.Database.getWritableDatabase(Database.java:200)
            at org.projectbuendia.client.sync.Database.clear(Database.java:185)
            at org.projectbuendia.client.widgets.EditAndClearDataPreference.onDialogClosed(EditAndClearDataPreference.java:32)
            at android.preference.DialogPreference.onDismiss(DialogPreference.java:391)
            at android.app.Dialog$ListenersHandler.handleMessage(Dialog.java:1271)
            at android.os.Handler.dispatchMessage(Handler.java:102)
            at android.os.Looper.loop(Looper.java:211)
            at android.app.ActivityThread.main(ActivityThread.java:5373)
            at java.lang.reflect.Method.invoke(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:372)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1020)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:815)
```

PR #109, opened 2015-10-30
Cursors use refactoring

llvasconcellos:
THIS IS NOT READY TO MERGE.
Investigating issue #107.
I'm trying to be sure all the cursors are being closed but the problem persists.
And as cursors are being used on an async task I think this might be the problem.
Anyways, I would like your input on this matter.
Cursors left open could generate memory leaks and other problems.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/models/AppModel.java
+++ b/app/src/main/java/org/projectbuendia/client/models/AppModel.java
@@ -301,22 +301,13 @@ public class AppModel {
         }
 
         @Override protected TypedCursor<T> doInBackground(Void... voids) {
-            Cursor cursor = null;
-            try {
-                cursor = mContentResolver.query(
-                    mContentUri,
-                    mProjection,
-                    mFilter.getSelectionString(),
-                    mFilter.getSelectionArgs(mConstraint),
-                    null);
-
+            try (Cursor cursor = mContentResolver.query(
+                mContentUri,
+                mProjection,
+                mFilter.getSelectionString(),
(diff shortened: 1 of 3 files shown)


llvasconcellos:
The problem is still happening. I just want to know if I'm on the right track.

app/src/main/java/org/projectbuendia/client/models/AppModel.java
```diff
@@ -301,22 +301,13 @@ public FetchTypedCursorAsyncTask(
...
-                    mFilter.getSelectionString(),
-                    mFilter.getSelectionArgs(mConstraint),
-                    null);
-
+            try (Cursor cursor = mContentResolver.query(
+                mContentUri,
+                mProjection,
+                mFilter.getSelectionString(),
+                mFilter.getSelectionArgs(mConstraint),
+                null)) {
```
llvasconcellos, line 309: I just noticed this change made the patient list stop working. Probably because it closes the cursor before returning it.
Is that right? How can we avoid this?

llvasconcellos, line 309: I think the problem that we are having is related to this TypedCursor because everywhere else I checked and cursors are being closed.

capnfabs, line 309: FYI, try-with-resources requires API level 19 (KitKat). I think we should use it for this project (where we control the hardware) but just wanted to let you know for the future. :)

The `try{} finally{c.close()}` pattern is what you should use if you need to support pre-KitKat.

Issue #27, opened 2015-09-24
Observation dialog - title set to "The page at 'file://' says:


dancunningham:
Title should be the name of the concept and the date.
e.g.
Oedema - 5 Sep

llvasconcellos:
This is the way chrome works when using alert() function. It's a security measure. The only way to avoid this is to create our own alert dialog.

capnfabs:
It's not too hard to fix though: http://stackoverflow.com/questions/10635563/modify-alert-title-javascript-in-android-webview

zestyping:
Wow, that's cool!  Maybe we could use that technique together with
Html.fromHtml to do the details popup in HTML as well, which would offer
lots more formatting flexibility.

capnfabs:
:+1:

G1DR4:
fixed, thx fabian.

PR #110, opened 2015-10-31
This fixes bug 27 (observation dialog titles)

G1DR4:
Bug 27 shows "File://" instead of a title for observation dialogs.
This fixes Bug 27 and is i18n compliant.
Changed all javascript alerts to android dialogs with customized title.

Still have to check possible API version problems

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
@@ -1,5 +1,9 @@
 package org.projectbuendia.client.ui.chart;
 
+import android.app.AlertDialog;
+import android.app.Activity;
+import android.content.Context;
+import android.content.DialogInterface;
 import android.content.res.Resources;
 import android.util.DisplayMetrics;
 import android.webkit.WebChromeClient;
@@ -45,7 +49,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 
 /** Renders a patient's chart to HTML displayed in a WebView. */
-public class ChartRenderer {
+public class ChartRenderer{
     static PebbleEngine sEngine;
     private static final Logger LOG = Logger.create();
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -45,7 +49,7 @@
 import java.util.TreeSet;
 
 /** Renders a patient's chart to HTML displayed in a WebView. */
-public class ChartRenderer {
+public class ChartRenderer{
```
llvasconcellos, line 52: You can leave the space before {

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -87,11 +91,33 @@ public void render(Chart chart, Map<String, Obs> latestObservations,
...
+                        .show();
+                return true;
+            }
+        });
+
         String html = new GridHtmlGenerator(chart, latestObservations, observations, orders,
                                             admissionDate, firstSymptomsDate).getHtml();
         mView.loadDataWithBaseURL("file:///android_asset/", html,
-            "text/html; charset=utf-8", "utf-8", null);
+                "text/html; charset=utf-8", "utf-8", null);
```
llvasconcellos, line 120: This auto indentation is a configuration in IntelliJ. You can import the settings file: android-studio-settings.jar to solve it.

G1DR4, line 120: thanks leo

llvasconcellos:
Rodrigo, I was testing and found some issues with this implementation. Not with what you did but the whole concept. I'll share my thoughts with you later.

llvasconcellos:
You can write somewhere in the commit message: "fixes #27" which will trigger github to automatically  close the issue.

llvasconcellos:
According to ping we should name our branches: feature/G1DR4-bla-bla-bla.

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -87,11 +91,33 @@ public void render(Chart chart, Map<String, Obs> latestObservations,
...
         mView.getSettings().setJavaScriptEnabled(true);
         mView.addJavascriptInterface(controllerInterface, "controller");
-        mView.setWebChromeClient(new WebChromeClient());
+
+        mView.setWebChromeClient(new WebChromeClient(){
+            @Override
+            public boolean onJsAlert(WebView view, String url, String message, final android.webkit.JsResult result)
+            {
+                new AlertDialog.Builder(mView.getContext())
+                        .setTitle(mView.getContext().getString(R.string.observation))
```
llvasconcellos, line 100: If you set the title as observation, every time a alert() is called in the chart it will have the title "Observation" even if it is not related. Just remove setTitle() and you will have a alert without title, like Firefox:
image

device-2015-11-04-070943

llvasconcellos:
So, the problem with this implementation is that does not have consistency with what we already have. 
This will assume that every alert() is an observation when it might be used somewhere else in the chart.
One other aspect is that does not attend @dancunningham needs on issue #27 when he says:

> Title should be the name of the concept and the date.
> e.g.
> Oedema - 5 Sep

To achieve such custom dialog you could use similar implementation did on orders that calls: controller.onOrderCellPressed which is PatientChartController's onOrderCellPressed().

llvasconcellos:
About API version It's working on 19 so let's worry about future compatibility issues later. Things might have change until then.

G1DR4:
i fix this in 'void observations'

PR #113, opened 2015-11-02
Refactor sync phase logic into individual sync phase controllers

capnfabs:
Before, there was a huge switch statement in SyncAdapter, which farmed out to individual methods,
which then almost always made a call to a static method on DbSyncUtil. This change pulls the logic
from the different "helper" methods into single, cohesive classes per sync phase.

Note that this commit is almost entirely moving code around, except for in
SyncAdapter#onPerformSync. SyncAdapter#onPerformSync is probably the only method worth reviewing in detail (it's difficult to review large-scale code moves). In my local testing it's worked so far, but I need to test with the automated tests before submitting.


capnfabs:
This is going to cause merge conflicts all over the place this sprint. I'll make a similar change after this sprint.

PR #114, opened 2015-11-02
Cleanup: Delete unused events and methods from UserManager

capnfabs:


viniciusboson:
Theses changes seem fine to me.

llvasconcellos:
@capnfabs I've found lots of unused code throughout the project. Is it ok to delete them as well?

llvasconcellos:
@capnfabs FunctionalTestCase.java has some. InstrumentationTestCase.java has one method too.

capnfabs:
@llvasconcellos it depends on the actual code. It's important to be especially careful with tests, because a commented out test often is a risk of regression, and it's better to make sure we know about it. But for unused code in the actual project, most of the time I think it's a good thing to do. If in doubt, feel free to send the PR and we can discuss.

PR #117, opened 2015-11-04
Logger now enforces a max TAG length of 23 chars.

capnfabs:
Android's Log.(whatever) methods throw IllegalArgumentExceptions for tags longer than 23 chars. This
change causes Logger to:
- Validate that user-specified tags are less than 23 chars at instantiation time, instead of when
  the Logger first tries to log (the previous behavior)
- Automatically truncate auto-generated tags to 23 chars.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/utils/Logger.java
+++ b/app/src/main/java/org/projectbuendia/client/utils/Logger.java
@@ -41,20 +41,30 @@ import org.projectbuendia.client.BuildConfig;
  */
 public final class Logger {
 
+    private static final int MAX_TAG_LENGTH = 23;
+
     public final String tag;
 
     /** Creates a {@link Logger} with the calling class's class name as a tag. */
-    public static final Logger create() {
+    public static Logger create() {
         return new Logger(getTag());
     }
 
-    private static final String getTag() {
+    private static String getTag() {
         String[] parts = new Throwable().getStackTrace()[2].getClassName().split("\\.");
-        return "buendia/" + parts[parts.length - 1];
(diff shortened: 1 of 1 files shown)


viniciusboson:
@capnfabs, could you explain me why 23 char length ?

app/src/main/java/org/projectbuendia/client/utils/Logger.java
```diff
@@ -41,20 +41,30 @@
...
+            return tag;
+        }
     }
 
     /** Creates a {@link Logger} with a manually-specified tag. */
-    public static final Logger create(String tag) {
+    public static Logger create(String tag) {
+        if (tag.length() > MAX_TAG_LENGTH) {
+            throw new IllegalArgumentException("Tag length should be less than " + MAX_TAG_LENGTH);
+        }
```
viniciusboson, line 67: Much better now, validating on creation.

capnfabs:
@viniciusboson Sure. Android Logging APIs throw IllegalArgumentExceptions if the tag is longer than 23 chars. Not sure why. Some more info here: http://stackoverflow.com/q/28168622/996592. I think I explained it a little in the commit message too.

viniciusboson:
yes, sure .. it's just up there. Thank you.
These changes look fine to me.

llvasconcellos:
@capnfabs where you having trouble with the tags or you just applied this fix to prevent future problems?

capnfabs:
I refactored some code and started having trouble with the tags after that. I ended up abandoning the refactor (see #113) because it would result in too many merge conflicts, but kept this commit.

PR #119, opened 2015-11-04
Fix UserManagerTest#testLoadKnownUsers_GeneratesEventOnFailure

capnfabs:
As of PR #91, having zero users is no longer an error. This test was previously based on
zero users being an error, so we change it such that the UserStore throws an error instead - a
condition that's still considered an error by the system.

### DIFF
--- a/app/src/androidTest/java/org/projectbuendia/client/user/UserManagerTest.java
+++ b/app/src/androidTest/java/org/projectbuendia/client/user/UserManagerTest.java
@@ -17,6 +17,8 @@ import com.google.common.collect.ImmutableSet;
 
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 import org.projectbuendia.client.FakeAsyncTaskRunner;
 import org.projectbuendia.client.events.user.KnownUsersLoadFailedEvent;
 import org.projectbuendia.client.events.user.KnownUsersLoadedEvent;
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/projectbuendia/client/user/UserManagerTest.java
```diff
@@ -52,19 +54,24 @@ public void testLoadKnownUsers_GeneratesEventOnSucess() throws Exception {
         mFakeAsyncTaskRunner.runUntilEmpty();
         // THEN the user manager fires off a KnownUsersLoadedEvent
         assertTrue(mFakeEventBus.getEventLog().contains(
-            new KnownUsersLoadedEvent(ImmutableSet.of(USER))));
+                new KnownUsersLoadedEvent(ImmutableSet.of(USER))));
```
viniciusboson: Probably this is a auto-indentation. It has 4 spaces more than it should has.

app/src/androidTest/java/org/projectbuendia/client/user/UserManagerTest.java
```diff
@@ -52,19 +54,24 @@ public void testLoadKnownUsers_GeneratesEventOnSucess() throws Exception {
...
+            }
+        });
         // WHEN loadKnownUsers is called and the async task is run
         mUserManager.loadKnownUsers();
         mFakeAsyncTaskRunner.runUntilEmpty();
         // THEN the user manager fires off a KnownUsersLoadFailedEvent
         mFakeEventBus.assertEventLogContains(
-            new KnownUsersLoadFailedEvent(KnownUsersLoadFailedEvent.REASON_NO_USERS_RETURNED));
+                new KnownUsersLoadFailedEvent(KnownUsersLoadFailedEvent.REASON_UNKNOWN));
     }
```
viniciusboson, line 75: Well done. I've realized yesterday that this method was not passing but i had no time fix it. 
 Probably this is a auto-indentation. It has 4 spaces more than it should has.

app/src/androidTest/java/org/projectbuendia/client/user/UserManagerTest.java
capnfabs: I can't figure out our code style :-/ Good catch though.

capnfabs:
Fixed both lint issues, please take a look.

viniciusboson:
It seems great to me.

llvasconcellos:
Tests passing! Good!

PR #120, opened 2015-11-04
Improve reliability of PatientChartActivityTest.

capnfabs:
Improve reliability of PatientChartActivityTest by waiting for views to be shown before attempting
to click on them.

I suspect the root cause might be something to do with the way our Activity is structured, or
general flakiness in Espresso, but this resulted in a vast improvement on my setup, pointed at the
dev server.

Note also that this doesn't fix the tests - there are multiple issues that should be addressed in
future commits - but this allows the tests to actually get to the assertions instead of failing in
the setup code.

### DIFF
--- a/app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
+++ b/app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
@@ -13,21 +13,15 @@ package org.projectbuendia.client.ui.chart;
 
 import android.support.test.espresso.Espresso;
 import android.support.test.espresso.IdlingResource;
+import android.support.test.espresso.ViewInteraction;
 import android.support.test.espresso.web.webdriver.Locator;
-
-import static android.support.test.espresso.web.assertion.WebViewAssertions.webMatches;
-import static android.support.test.espresso.web.sugar.Web.onWebView;
-import static android.support.test.espresso.web.webdriver.DriverAtoms.findElement;
-import static android.support.test.espresso.web.webdriver.DriverAtoms.getText;
-
-import static org.hamcrest.Matchers.containsString;
-
 import android.test.suitebuilder.annotation.MediumTest;
 import android.view.View;
 import android.widget.CheckBox;
 import android.widget.EditText;
 import android.widget.RadioButton;
 
+import org.hamcrest.Matcher;
 import org.odk.collect.android.views.MediaLayout;
 import org.odk.collect.android.views.ODKView;
 import org.odk.collect.android.widgets2.group.TableWidgetGroup;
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
```diff
@@ -51,6 +58,9 @@
     private static final String YES = "●";
 
     private static final int ROW_HEIGHT = 84;
+    private static final Matcher<View> OVERFLOW_BUTTON_MATCHER = anyOf(
+            allOf(isDisplayed(), withContentDescription("More options")),
+            allOf(isDisplayed(), withClassName(endsWith("OverflowMenuButton"))));;
```
viniciusboson, line 64: Probably this is a auto-indentation. It has 8 spaces more than it should has.

capnfabs, line 64: Actually I think this is ok. We use 8 spaces for line continuation. See https://github.com/projectbuendia/buendia/wiki/Android-Style-Guide#452-continuation-indentation-at-least-8-spaces

viniciusboson, line 64: humm, i see ... Sorry about that.

viniciusboson, line 64: I haven't seen this Android Style Guide before, just the java Style.
I've updated our wiki side bar to link to this page too.

viniciusboson:
this change looks fine to me.

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
G1DR4, line 64: dont worry about the wiki. it's going to change

G1DR4:
it's fine to me too.

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
```diff
@@ -148,13 +158,15 @@ public void testPatientChart_CanOpenEncounterFormMultipleTimes() {
...
 
         EventBusIdlingResource<FetchXformSucceededEvent> xformIdlingResource =
-            new EventBusIdlingResource<FetchXformSucceededEvent>(
-                UUID.randomUUID().toString(), mEventBus);
-        click(viewWithText("[test] Form"));
+                new EventBusIdlingResource<>(UUID.randomUUID().toString(), mEventBus);
+        ViewInteraction testForm = viewWithText("[test] Form");
+        expectVisibleSoon(testForm);
+        click(testForm);
         Espresso.registerIdlingResources(xformIdlingResource);
```
llvasconcellos, line 170: The above lines are great! I tried to fix the problem of "[test] Form" not found but my limited knowledge of android / spresso stopped me from doing it.
I just tested it and it did not happen once, so I summon your mighty force to repeat this with the webview idling resource. :) (of course, when you have time). It is another thing I have an branch but did not have success implementing.

llvasconcellos:
Just on last comment though, I know this is not related to this PR but I noticed some of the tests that was passing aren't anymore. I'm scared! 
@capnfabs I was going to open a new issue for it but I see that you already did it.
The rounding problem was already fixed but some PR might have overridden it or I maybe it's just a format in the profile csv. I'll have to search our closed issues and PRs to find out.
I'll get back to you about that.

capnfabs:
I think the rounding thing was the only other problem in the tests. Thanks for looking into it! I imagine we can fix the problem by changing the formatting in the profile to whatever the normal observation fields use.

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
capnfabs, line 170: I don't know much about idling resources either, which is why I added these `expectVisibleSoon` calls. I've got a feeling that our tests aren't doing it correctly at the moment though if the idling resource detection logic is failing.

PR #124, opened 2015-11-05
Make PatientChartActivityTest slightly less flaky.

capnfabs:
Previously, testEncounter_latestEncounterIsAlwaysShown would take 3 minutes to run because there was
a 30 second wait and 6 iterations. A 5 second wait is enough for it to work, and 3 iterations is
enough to actually test the same behavior.

Fix an error in testCombinesNonOverlappingObservationsForSameEncounter such that it works with the
test profile and with the dev server.

### DIFF
--- a/app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
+++ b/app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
@@ -252,7 +252,7 @@ public class PatientChartActivityTest extends FunctionalTestCase {
 
         // Update a vital tile (pulse) as well as a couple of observations (temperature, vomiting
         // count), and verify that the latest value is visible for each.
-        for (int i = 0; i < 6; i++) {
+        for (int i = 0; i < 3; i++) {
             openEncounterForm();
 
             String temp = Integer.toString(i + 35) + ".7";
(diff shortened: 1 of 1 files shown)


llvasconcellos:
I added 30 seconds because at the time we had 700+ patients and 50+ users on dev.projectbuendia.org and it was really, really slow.

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
```diff
@@ -352,7 +352,11 @@ public void testCombinesNonOverlappingObservationsForSameEncounter() {
         waitForProgressFragment();
         // Expect a WebView with JS enabled to be visible soon (the chart).
         expectVisibleSoon(viewThat(isJavascriptEnabled()));
-        checkObservationValueEquals("[test] Temperature (°C)", "36");
+        try {
+            Thread.sleep(5000);
+        } catch (InterruptedException ignored) {
+        }
+        checkObservationValueEquals("[test] Temperature (°C)", "36.5");
```
llvasconcellos, line 359: Good!

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
```diff
@@ -271,8 +271,8 @@ public void testEncounter_latestEncounterIsAlwaysShown() {
             // TODO: implement IdlingResource for webview to remove this sleep.
             // Wait a bit for the chart to update it's values.
             try{
-                Thread.sleep(30000);
-            } catch (InterruptedException e){}
+                Thread.sleep(5000);
+            } catch (InterruptedException ignored){}
```
llvasconcellos, line 275: I like the "ignored" exception name. :+1:

llvasconcellos:
I won't even test this. Seems good to me.

app/src/androidTest/java/org/projectbuendia/client/ui/chart/PatientChartActivityTest.java
capnfabs, line 275: That's got a specific meaning - the linter in Android Studio flags empty catch blocks, unless the exception is deliberately named "ignored".

capnfabs:
Thanks for the context re: 30 seconds; hopefully we won't have to increase it again :)

PR #126, opened 2015-11-05
Rename Gender --> Sex as per TODO

capnfabs:


llvasconcellos:
@capnfabs Why rename?

llvasconcellos:
This change is good to go!

capnfabs:
As per the comment in the corresponding code in the server, OpenMRS uses the term "gender" when it actually means "sex". That doesn't mean that we have to, though. It was an outstanding TODO that I decided to fix.

PR #129, opened 2015-11-06
Generalize sync token storage.

capnfabs:
Previously, a server-supplied timestamp was stored for each incremental observation sync. This
timestamp was stored in a field in the "misc" SQLite table.

There are two problems with this:
- Our design (http://j.mp/buendia-sync) advocates the client treating the server-supplied token as
  an opaque sync token instead of as a timestamp
- Soon we'll be storing sync tokens for patients as well.

This change refactors the incremental observation timestamp storage into a more general "sync token"
storage, which can be used by multiple sync phases.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/providers/Contracts.java
+++ b/app/src/main/java/org/projectbuendia/client/providers/Contracts.java
@@ -154,7 +154,7 @@ public class Contracts {
     public interface SyncTokens {
         Uri CONTENT_URI = buildContentUri("sync-tokens");
         String ITEM_CONTENT_TYPE = buildItemType("sync-token");
-        String TYPE = "type";
+        String TABLE_NAME = "table_name";
         String SYNC_TOKEN = "sync_token";
     }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/projectbuendia/client/providers/Contracts.java
```diff
@@ -148,14 +149,13 @@ public String toString() {
...
-         * <p/>
-         * <p>Updated after observation sync to the snapshot time reported by the server.
-         */
-        String OBS_SYNC_END_MILLIS = "obs_sync_end_millis";
+    }
+
+    public interface SyncTokens {
+        Uri CONTENT_URI = buildContentUri("sync-tokens");
+        String ITEM_CONTENT_TYPE = buildItemType("sync-token");
+        String TYPE = "type";
```
zestyping: Overall this is great!

The `type` column is a bit vague though.  Can we just say that this is a table name?  That would both make clear what goes in this column and let us reuse the Table enum instead of adding a duplicate OBSERVATIONS constant.

capnfabs: Good call. It's much better with `table_name` instead of type, and I like the additional type-safety in the `SyncAdapter#[store/get]SyncToken` methods.

PR #128, opened 2015-11-05
Fixing error when OdkDatabase tries to fetch the form id from a nonexistent file

viniciusboson:
The method `OdkDatabe#getFormIdForPath` was throwing an `CursorIndexOutOfBoundsException` when it was trying to fetch the form id from a nonexistent file. But the proper was, as documented in javadoc, was to return `-1` if no match was found.

Here the full stack trace

``` java

android.database.CursorIndexOutOfBoundsException: Index 0 requested, with a size of 0
at android.database.AbstractCursor.checkPosition(AbstractCursor.java:426)
at android.database.AbstractWindowedCursor.checkPosition(AbstractWindowedCursor.java:136)
at android.database.AbstractWindowedCursor.getLong(AbstractWindowedCursor.java:74)
at android.database.CursorWrapper.getLong(CursorWrapper.java:106)
at org.projectbuendia.client.net.OdkDatabase.getFormIdForPath(OdkDatabase.java:50)
at org.projectbuendia.client.net.OdkDatabaseTest.testInexistentFile_getFormIdForPathMustReturnOneNegative(OdkDatabaseTest.java:30)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at android.test.InstrumentationTestCase.runMethod(InstrumentationTestCase.java:214)
at android.test.InstrumentationTestCase.runTest(InstrumentationTestCase.java:199)
at junit.framework.TestCase.runBare(TestCase.java:134)
at junit.framework.TestResult$1.protect(TestResult.java:115)
at junit.framework.TestResult.runProtected(TestResult.java:133)
at android.support.test.internal.runner.junit3.DelegatingTestResult.runProtected(DelegatingTestResult.java:90)
at junit.framework.TestResult.run(TestResult.java:118)
at android.support.test.internal.runner.junit3.AndroidTestResult.run(AndroidTestResult.java:49)
at junit.framework.TestCase.run(TestCase.java:124)
at android.support.test.internal.runner.junit3.NonLeakyTestSuite$NonLeakyTest.run(NonLeakyTestSuite.java:63)
at junit.framework.TestSuite.runTest(TestSuite.java:243)
at junit.framework.TestSuite.run(TestSuite.java:238)
at android.support.test.internal.runner.junit3.DelegatingTestSuite.run(DelegatingTestSuite.java:103)
at android.support.test.internal.runner.junit3.AndroidTestSuite.run(AndroidTestSuite.java:63)
[...]

PR #130, opened 2015-11-07
Open form offline

viniciusboson:
These changes were made to allow open any form offline.

When the application synchronizes with the server initially (`SynchAdapter`), now it fetches all the forms from the server and caches them. Given that most of the time it will not have many forms, this change will not impact on synch performance.

Then, whenever the application tries to access a form, it tries to load the cached form first, and if it is not available, then the apps tries to fetch the form from the server and to cache it. If there is no such neither locally or on server, then an error event is triggered. 

As @dancunningham said in one of his comments bellow, " I think this behaviour is better as it will make the opening of forms more responsive and at the moment we saw in field testing that delay of a few seconds on opening a form - just when you want to add data - is annoying". 

The changes on `OdkDatabase.java` and `OdkDatabaseTest.java` files are actually on PR #128, but as it is not merged yet, it appears here.


app/src/main/java/org/projectbuendia/client/sync/SyncAdapter.java
```diff
@@ -267,6 +268,7 @@ public SyncAdapter(Context context, boolean autoInitialize, boolean allowParalle
                     // there are only a few forms, usually less than 10.
                     case SYNC_FORMS:
                         updateForms(provider, syncResult);
+                        OdkActivityLauncher.fetchAndCacheAllXforms();
```
viniciusboson, line 271: Here i was in doubt if `OdkActivityLauncher.fetchAndCacheAllXforms();` should be inside the `updateForms` method.
But i' rather to keep them isolated given they do different things.

app/src/main/java/org/projectbuendia/client/net/OdkDatabase.java
```diff
@@ -36,18 +36,21 @@ public static long getFormIdForPath(File path) {
...
+            //if any form was found by the given file.
+            if(cursor.getCount() >= 1) {
+                // There should only ever be one form per UUID. But if something goes wrong, we want
+                // the app to keep working. Assume the latest one is correct.
+                if (cursor.getCount() > 1) {
+                    Log.e(TAG, "More than one form in database with the same id. This indicates an "
+                        + "error occurred on insert (probably a race condition) and should be "
+                        + "fixed. However, the app should still function correctly");
+                }
+                Preconditions.checkArgument(cursor.getColumnCount() == 1);
```
zestyping: This check is redundant; the projection in the query is just 10 lines up and there's only one column.

app/src/main/java/org/projectbuendia/client/net/OdkXformSyncTask.java
```diff
@@ -135,27 +112,39 @@ private static boolean writeStringToFile(String response, File proposedPath) {
...
-                }
-            });
+            // Doesn't exist, so insert it. Fetch the file from OpenMRS
+            fetchAndAddXFormToDb(formInfo.uuid, proposedPath);
         }
         return null;
     }
 
     /**
+     * Fetches the request xforms from the server and adds it into db.
```
zestyping: fetches one form

viniciusboson:
Thx for reviewing it, @zestyping. I corrected what you suggested me..

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -111,40 +142,30 @@ public static void fetchAndShowXform(
...
-                            LOG.i("wrote form " + path);
-                            showOdkCollect(
-                                callingActivity,
-                                requestCode,
-                                OdkDatabase.getFormIdForPath(path),
-                                patient,
-                                fields);
-                        }
-                    }).execute(findUuid(response, uuidToShow));
+                    // Cache the form into the ODK form cache
```
capnfabs: I think this comment is out of date now.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -180,6 +201,78 @@ public static void showOdkCollect(
         callingActivity.startActivityForResult(intent, requestCode);
     }
 
+    private static List<OpenMrsXformIndexEntry> getLocalFormEntries() {
+        List<OpenMrsXformIndexEntry> entries = new ArrayList<>();
+
+        final ContentResolver resolver = App.getInstance().getContentResolver();
+        Cursor c = resolver.query(Contracts.Forms.CONTENT_URI, null, null, null, null);
```
capnfabs: It's generally better to specify the projection (the columns) for database reads, in case the database schema changes in the future.

capnfabs:
Thanks for doing this @viniciusboson. Two high-level questions:
- Why do we need to change the code that syncs users in order to make sure that forms are usable offline?
- If we're syncing all the forms regularly, do we need to try to fetch the form from the server every time the user loads a form?

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson: Thx @capnfabs. I deleted this comment. The `showForm`'s javacdoc was out of date too. I corrected it and removed an unused param too.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson: Thx @capnfabs. I specified the projection here and from the source method `DbSyncHelper#getFormUpdateOps` where i've copied it.

viniciusboson:
Good questions, @capnfabs ...
Actually i've changed the code that syncs users because i initially thought that it would be necessary to load the user list offline too. Then, i realized that it already loads offline. But as i've just overloaded the method but i didn't change any behavior from the its callers, not causing any side effect, i thought that this change would remain in this PR. If you think, it should be in another PR, i can extract it from this one.

About trying to fetch the form from the server each time it is accessed, it was just a manner to guarantees that the user is using the updated form. But, in fact, considering that it synchs the form regularly and the form isn't changed so often, this double check doesn't have great value. I haven't thought about that but now i agree with you.

viniciusboson:
@capnfabs, 
Considering that the application synchs frequently and the forms don't change often, i've changed the code to when the user tries to load a form, the app checks if it is cached. If so, it loads locally. Otherwise it fetches from the server.

app/src/main/java/org/projectbuendia/client/sync/SyncAdapter.java
capnfabs, line 271: Sounds good, keep them separate. :+1:

capnfabs:
Thanks @viniciusboson. It would be better to move the user syncing changes to a separate PR, it keeps the scope of both reviews smaller and more focussed.

capnfabs:
Regarding whether it's better to attempt a form fetch when the user loads the form or rely upon the synced behaviour, I'm honestly not sure about which is better. Does anybody else have any opinions on this?

app/build.gradle
```diff
@@ -137,7 +137,7 @@ def serverEdison = 'server';
 def serverEmulatorHost = '10.0.2.2';
 
 // Server on the Fantastica wifi network.
-def serverFantastica = '192.168.1.161';
+def serverFantastica = '192.168.1.179';
```
capnfabs: Is this change something that should be checked in? I have no idea what this WiFi network is :)

dancunningham:
Let's see if this Pivotal Tracker integration works, mentioning a Tracker story ID in square brackets like this [#107200340] should make this comment appear against the story in Tracker too. It might only work on commit messages though, rather than comments on pull requests...

app/build.gradle
viniciusboson: ouw, my bad !!! I was not supposed to commit it. I will fix it later.

viniciusboson:
Great, @dancunningham. I'll try Pivotal tracker on my next commit. :+1:

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -180,6 +188,105 @@ public static void showOdkCollect(
...
+                entries.add(new OpenMrsXformIndexEntry(uuid, name, date));
+            }
+        } finally {
+            c.close();
+        }
+
+        return entries;
+    }
+
+    private static void handleSyncError(VolleyError error) {
```
capnfabs, line 239: Suggest renaming this to `handleFetchError`, because we could be fetching a form when we're not syncing.

capnfabs:
This looks great! OK to merge after addressing other comments. 👌 Please make sure you test extensively before merging, this changes some very important behaviour.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -85,7 +85,37 @@
     private static final Logger LOG = Logger.create();
 
     /**
-     * Fetches all xforms from the server, caches them, and launches ODK using the requested form.
+     * Fetches all xforms from the server and caches them. If any error occurs during fetching,
+     * a failed event is triggered.
+     */
+    public static void fetchAndCacheAllXforms() {
+        new OpenMrsXformsConnection(App.getConnectionDetails()).listXforms(
```
dancunningham, line 92: Just want to check - this only retrieves the "published" forms, right? Not all of the forms? (there will be a number of forms in OpenMRS that are not published as they were created for other profiles)

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -102,6 +132,11 @@ public static void fetchAndShowXform(
         final int requestCode,
         @Nullable final org.odk.collect.android.model.Patient patient,
         @Nullable final Preset fields) {
+        LOG.i("Trying to fetch it from cache.");
```
dancunningham, line 135: This appears to be the other way around from how you describe in the PR description.

It looks like it tries the cache first. If it succeeds in getting it from the local cache, it returns. If it doesn't, it tries to fetch it from the server.

That does sound like it will be faster as it doesn't have to go to the server each time.

The implication is that if a form is changed, the new version will never be retrieved unless that cache is invalidated. Therefore we probably need some way to clear the forms cache and force it to get the forms from the server (maybe in the settings).

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -111,40 +146,13 @@ public static void fetchAndShowXform(
                             FetchXformFailedEvent.Reason.NO_FORMS_FOUND));
                         return;
                     }
-                    // Cache all the forms into the ODK form cache
```
dancunningham, line 114: Good that we're no longer doing this every time!

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -85,7 +85,37 @@
...
+        new OdkXformSyncTask(null).fetchAndAddXFormToDb(formEntry.uuid,
+            formEntry.makeFileForForm());
+    }
+
+    /**
+     * Loads the xform from the cache and launches ODK using it. If the cache is not available,
+     * the app tries to fetch it from the server. If no form is got, it is triggered a failed event.
      * @param callingActivity the {@link Activity} requesting the xform; when ODK closes, the user
      *                        will be returned to this activity
      * @param uuidToShow      UUID of the form to show
```
dancunningham, line 121: This could be better named e.g. formUuid

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
dancunningham, line 135: @viniciusboson - I think this behaviour is better as it will make the opening of forms more responsive and at the moment we saw in field testing that delay of a few seconds on opening a form - just when you want to add data - is annoying.

Please update the PR description to reflect what you are actually doing.

Now - we do need a way to invalidate the cache and force retrieval of forms from the server. How might we do that? Maybe add a "Clear forms cache" option in our App Settings.

Later, we'll need to think about how that will work with deployment of new forms from HQ to field projects.

capnfabs, line 135: @dancunningham I think what we want is to force-refresh the forms on app-wide sync. It's more bullet-proof than a setting. What do you think?

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson, line 92: Do you mean the only the current profile's forms ? If so, yeap @dancunningham .

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson, line 135: Sorry about that, @dancunningham. I completely forgot to update the PR description. 
Initially, it was fetching from the server first, but after discussing with @capnfabs, we've realized that this was not necessary because now the form are synched frequently, together with other datas, and a profile change is much less frequent.  So, i agree that the apps should priorize the cached form. To invalidade the cache, you just need to go to settings and invalidade all the other datas. This is actually the big point. The forms work as the other datas. You don't need to check anywhere that you are working on offline mode.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson, line 121: Sounds much better, @dancunningham

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
dancunningham, line 92: Yes, good!

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
dancunningham, line 135: @capnfabs when is an app-wide sync triggered?

If we have 10,000 patients, would it be unnecessarily slow to do an app-wide sync just to get updated forms? Or not, now that we have incremental sync?

dancunningham:
This looks good. I'll merge it into wip to trigger a Jenkins build and test this on the tablet today, before we merge this into dev.

dancunningham:
I notice that if I make a change to a form on OpenMRS, then do a pull-down refresh in the app, then disconnect wi-fi, then it does successfully pick up the form change. This is great!

However, if I publish a new form, then do a pull-down refresh in the app, then disconnect wi-fi, it has not picked up that new form.

It seems that for a new form to be picked up, it requires re-doing the initial sync. i.e. going in to the settings and hitting "Apply and clear local data".

It would be good if new forms were also picked up by the incremental sync.

[raised this as an issue - https://github.com/projectbuendia/client/issues/148]

dancunningham:
This looks like it is working well, I'm going to merge it!

PR #133, opened 2015-11-09
Tidy up `Gson` serializers/deserializers.

capnfabs:
- Removed duplicate serializers (for Joda `DateTime`), renamed `DateTimeDeserializer` to `DateTimeSerializer`.
- Moved `NetModule`'s `Gson` provider method to use `Serializers#registerTo` class, which makes all usages consistent across the app.

Ideally, we'd move the things that aren't using the injected Gson to using the injected one, but that's a larger change, so we'll put it off.


llvasconcellos:
No comments. Good to go.

PR #137, opened 2015-11-10
Refactoring UserStore to be more readable.

viniciusboson:
These changes are not supposed to cause any side effect on the current code.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/user/UserStore.java
+++ b/app/src/main/java/org/projectbuendia/client/user/UserStore.java
@@ -21,6 +21,7 @@ import android.os.RemoteException;
 import com.android.volley.Response;
 import com.android.volley.VolleyError;
 import com.android.volley.toolbox.RequestFuture;
+import com.google.common.collect.ImmutableSet;
 
 import net.sqlcipher.database.SQLiteException;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/projectbuendia/client/user/UserStore.java
```diff
@@ -45,90 +46,61 @@
...
+    /** Adds a new user, both locally and on the server. */
+    public JsonUser addUser(JsonNewUser user) throws VolleyError {
+        JsonUser newUser = addUserOnServer(user);
+        addUserLocally(newUser);
+        return newUser;
+    }
 
-        LOG.i("Got %d users from server; updating local database", users.size());
+    /** Deletes a user, both locally and on the server. */
+    public JsonUser deleteUser(JsonUser user) {
```
capnfabs: You can probably delete this method while you're here :)

capnfabs:
This looks good! OK to merge after deleting the unused method. Thanks!

dancunningham:
@viniciusboson can you make the change that @capnfabs suggested and then merge this please!

app/src/main/java/org/projectbuendia/client/user/UserStore.java
viniciusboson: @capnfabs. I was wondering about that. I decided to keep it to explicit state that you can't delete an user, rather then you just think that it is not implemented yet.
But let's keep the code clean. It is deleted now.

capnfabs: Thanks! Ok to merge.

In general, it's better to delete unused code because it makes the process of changing code in the future simpler.

viniciusboson:
Merged :+1:

app/src/main/java/org/projectbuendia/client/user/UserStore.java
viniciusboson: Yes, it makes much sense.

PR #138, opened 2015-11-10
Error response treatment

llvasconcellos:
Although it's a work in progress this PR is ready to merge.


llvasconcellos:
To see the benefit of JsonErrorListener I'll still have to add different behaviors depending on the error. Now, for each error it simply output the message. This might be not a desired behavior depending on the task being done.

llvasconcellos:
device-2015-11-10-053640

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
```diff
@@ -0,0 +1,70 @@
...
+import com.android.volley.Response.ErrorListener;
+import com.android.volley.VolleyError;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonParser;
+
+import org.projectbuendia.client.App;
+
+public class JsonErrorListener implements ErrorListener {
```
capnfabs: It's a bit strange to have a network request updating UI. I'm not sure what the best way to solve this is, but as an idea / starting point, maybe broadcast something to the event bus from this network request? Or do you think that's just obfuscating the real behaviour?

capnfabs:
I think your general approach for this is good. It's definitely good to have utility functionality that can "understand" OpenMRS messages and convert them into a response - but I think you're right that we're going to want situation-specific behaviour for lots of cases.

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
```diff
@@ -0,0 +1,70 @@
...
+                    if (element == null || element.isJsonNull()) {
+                        element = errorObject.get("code");
+                    }
+                    if (element != null && element.isJsonPrimitive()) {
+                        message = element.getAsString();
+                    }
+                }
+            }
+        } catch (JsonParseException | IllegalStateException | UnsupportedOperationException e) {
+            e.printStackTrace();
```
capnfabs: Maybe use `LOG.w` here instead.

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
capnfabs: Maybe rename this to something that indicates that errors get turned into toasts. Maybe `ToastErrorListener`?

capnfabs:
Ok to merge after responding to in-line comments. :+1: Thanks @llvasconcellos!

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
llvasconcellos: @capnfabs I added a TODO to refactor the message using eventbus and snackbar.
I renamed the class to OpenMrsErrorListener.

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
llvasconcellos: Done.

app/src/main/java/org/projectbuendia/client/net/JsonErrorListener.java
llvasconcellos: Just one thing: I did not rename it ToastErrorListener because it is not the only thing it could do. If you override the onErrorResponse method when making your request you could alter it's behavior but using the utility methods of the class.

llvasconcellos:
Now it shows non Json error as well.

PR #139, opened 2015-11-10
Switch to incrementally syncing observations instead of encounters.

capnfabs:
See corresponding PR https://github.com/projectbuendia/buendia/pull/30 for rationale.


G1DR4:
Good to go!!

PR #141, opened 2015-11-11
Tidy up the build.gradle file for the app.

capnfabs:
- Remove an unused dependency, and a mysterious dependency that isn't required
- Add comments describing what some of the libraries are for.


capnfabs:
Rolled this change into a bigger PR.

Issue #144, opened 2015-11-12
Sync - delay after completing form before results appear in chart


dancunningham:
I completed the Intensive Observations form and pressed Save.

It took 60s before the observations appeared in the chart.

This used to happen locally immediately; not needing to wait for a sync.

It should happen immediately so the nurse can verify that the observations have gone in correctly.

capnfabs:
Good catch. The problem is that I changed the database schema to include Observation UUID and make it compulsory, but when we insert values locally to ensure immediate updates, we don't add a UUID.

The solution is to make the UUID field optional.

capnfabs:
PR out: https://github.com/projectbuendia/client/pull/145

PR #145, opened 2015-11-13
Fix Issue #144

capnfabs:
- Relax `NOT NULL` constraint on local Observation UUIDs.
- Remove `temp_cache` database column in favor of using NULL uuids for locally inserted observations.


dancunningham:
Tested this by merging it into wip branch and testing the build.

Looks good! Nice and responsive now, chart updates a couple seconds after filling in the form. Works for backdated observations. Tiles and grid updates. Great!

PR #146, opened 2015-11-13
Refactor SyncAdapter.

capnfabs:
`SyncAdapter` is a really important part of our app, but it's gotten a little out of hand - the logic for each sync phase is spread across a bunch of "helper" methods (`DbSyncUtil`, I'm looking at you :angry:).

This change:
- Moves the various `updateFoo` methods and their counterparts into separate, phase-specific, `SyncPhaseControllers`.
- Improves the way sync progress is reported.

There are two big goals of this refactor:
- _Readability_: Make each sync phase easier to work on and maintain, because logic for a single sync phase is contained in a single, better-contained file
- _Enable future code reuse_: Open up the possibility of factoring out common incremental sync handling code in the future.

Review notes:
- I'd suggest reviewing this code commit-by-commit, I've designed each commit to be self-contained and relatively straightforward. 
- Most of the delta is just moving lines of code between files. The only things that aren't moving code around are:
  - I've removed the occasional `checkCancellation()` method call - cancellation is only checked between phases now.
  - The contents of commit e8ab6ab.
- I've put these all in a single PR because they all build on each other, it's a single refactor and it's extremely difficult to structure multiple PRs like this. Having said that, let me know if you want me to factor any out into separate PRs, especially commit e8ab6ab.


app/src/main/java/org/projectbuendia/client/sync/SyncAdapter.java
```diff
@@ -103,14 +69,23 @@
...
     public enum SyncPhase {
-        SYNC_USERS,
-        SYNC_LOCATIONS,
-        SYNC_CHART_ITEMS,
-        SYNC_CONCEPTS,
-        SYNC_PATIENTS,
-        SYNC_OBSERVATIONS,
-        SYNC_ORDERS,
-        SYNC_FORMS
+        SYNC_USERS(R.string.syncing_users, new UsersSyncPhaseController()),
```
zestyping: This is great!  So nice to see each phase in its own class.

Why not go all the way and make them all SyncPhases?  SyncPhases and SyncPhaseControllers don't need to be two separate things; couldn't they be a single interface? As in:

```
interface SyncPhase {
    String getName();
    String getMessage();
    void sync();
}
```

capnfabs: Thanks!

Under your proposal, how does client code request that only a subset of phases are synced? Would you keep the enum for that, or have a different mechanism?

capnfabs: I think the phase name and message are handled better at the enum level. By keeping `SyncPhaseController` as a purely functional object it makes it easier to composite logic later.

So my planned next step in this direction is something like have an `IncrementalSyncPhaseController<T> implements SyncPhaseController` that can be used to wrap another function which allows us to generalize incremental sync logic. I think that's cleaner than the approach we're currently using, which is to have helper method calls at the start and end of the `sync()` function for those `SyncPhaseControllers` that do incremental sync.

llvasconcellos:
I made some minor comments but it's good to go.

app/src/main/java/org/projectbuendia/client/sync/SyncAdapter.java
zestyping: To answer your question: each SyncPhase has a name returned by getName(), and the SyncAdapter has a Map from names to SyncPhase instances.  The caller requests a selection of phases by passing a bundle of flags, just as it does now, specifying the names of the phases.

In part my proposal is motivated by simplicity, and in part motivated by an inappropriate use of the name "Controller" -- these aren't controllers in the MVC sense, and the addition of "controller" to the name doesn't tell us any more information about what they are.

The interface seems to fit reasonably well -- why wouldn't incremental sync phases also have names and status messages?

capnfabs: I agree with you about the `Controller` thing - it's a filler word that I used because I didn't have a better word. In its current form, this is more of a `SyncPhaseRunnable`. I will rename it to that.

I think it's better as a narrower interface though - the existing enum handles the registration for us, and I think it's good to keep meta-information about the sync phases separate from the logic of the sync phases.

capnfabs:
Wow, that merge was a doozy. Tested it by:
- running the tests (all passing)
- manually checking the behaviour of both conflicting changes.

PR #147, opened 2015-11-13
Remove obsolete developer settings.

capnfabs:
Remove "Update local observations on submission" and "Sync observations incrementally" settings.

These settings have been 'on' by default for a long time, and we're relying on the behaviour of both
of them extensively.

Also:
- rename some methods named `startIncrementalObsSync` to `startObservationsSync`, because all
  observations syncs are incremental syncs now.
- delete unused `OdkActivityLauncher#showSavedXform()` method.

### DIFF
--- a/app/src/main/res/xml/pref_developer.xml
+++ b/app/src/main/res/xml/pref_developer.xml
@@ -9,20 +9,6 @@
         android:summary="@string/pref_desc_store_form_instances"
         android:defaultValue="@bool/keep_form_instances_locally_default" />
 
-    <!-- Whether to instantly update the local cache when an xform is submitted locally -->
-    <CheckBoxPreference
-        android:key="xform_update_client_cache"
-        android:title="@string/pref_title_xform_update_client_cache"
-        android:summary="@string/pref_desc_xform_update_client_cache"
-        android:defaultValue="@bool/xform_update_client_cache_default" />
-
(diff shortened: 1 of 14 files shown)


app/src/main/res/xml/pref_developer.xml
```diff
@@ -9,20 +9,6 @@
...
-        android:summary="@string/pref_desc_xform_update_client_cache"
-        android:defaultValue="@bool/xform_update_client_cache_default" />
-
-    <!-- Whether to instantly update the observations cache incrementally -->
-    <CheckBoxPreference
-        android:key="incremental_observation_update"
-        android:title="@string/pref_title_incremental_observation_update"
-        android:summary="@string/pref_desc_incremental_observation_update"
-        android:defaultValue="@bool/incremental_observation_update_default" />
-
```
llvasconcellos, line 25: Aren't you going to remove the strings as well?

llvasconcellos:
Good to merge after removing unused strings.

app/src/main/res/xml/pref_developer.xml
capnfabs, line 25: Good catch. Done.

G1DR4:
good to go too

app/src/main/res/xml/pref_developer.xml
llvasconcellos, line 25: :thumbsup:

PR #150, opened 2015-11-18
This fixes #106 (Stop health check when wrong password)

### LINKED ISSUE #106: Incorrect password causes lockout
OpenMRS locks out a user for 5 minutes after 7 incorrect login attempts in a row. 

https://wiki.openmrs.org/display/docs/Administering+Users#AdministeringUsers-ManagingUserLockout

It looks like the client app keeps trying to log in even when it's getting back an "incorrect password" response.

This is on a per-user database so when all apps are sharing a user it's a lot more likely to happen.

Because the tablets keep trying periodically to log in they automatically hit the limit and get locked out.

Most appropriate fix is probably if the tablet sees it is getting an incorrect password response, it should stop trying until you edit the settings and hit "Apply" on either the server, username or password dialog.

To workaround this for now and unlock the user, in mysql I had to run:

```
delete from user_property where user_id=4 and property=‘lockoutTimestamp’;
delete from user_property where user_id=4 and property='loginAttempts’;
```

llvasconcellos:
Stop health check when wrong password.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -161,6 +161,8 @@ def requireWifi = project.requireWifi ?: 'false';
 // database unencrypted so that it can be easily inspected using Stetho.
 def encryptionPassword = project.encryptionPassword ?: '';
 
+def syncAccountName = "sync";
+
 logger.info('Default server hostname: ${server}')
 logger.info("Default OpenMRS root URL: ${openmrsRootUrl}")
 logger.info("Default OpenMRS user: ${openmrsUser}")
@@ -208,6 +210,7 @@ android {
         buildConfigField 'String', 'CONTENT_AUTHORITY', '"' + contentAuthority + '"'
         resValue 'string', 'account_type', appId
         buildConfigField 'String', 'ACCOUNT_TYPE', '"' + appId + '"'
+        buildConfigField 'String', 'ACCOUNT_NAME', '"' + syncAccountName + '"'
 
         // Set the SQLite database encryption password.
         buildConfigField 'String', 'ENCRYPTION_PASSWORD', '"' + encryptionPassword + '"'
--- a/app/src/main/java/org/projectbuendia/client/diagnostics/BuendiaApiHealthCheck.java
+++ b/app/src/main/java/org/projectbuendia/client/diagnostics/BuendiaApiHealthCheck.java
@@ -158,7 +158,7 @@ public class BuendiaApiHealthCheck extends HealthCheck {
                             case HttpURLConnection.HTTP_FORBIDDEN:
                             case HttpURLConnection.HTTP_UNAUTHORIZED:
                                 reportIssue(HealthIssue.SERVER_AUTHENTICATION_ISSUE);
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/projectbuendia/client/diagnostics/BuendiaApiHealthCheck.java
```diff
@@ -158,6 +158,7 @@ public BuendiaModuleHealthCheckRunnable(Handler handler) {
                             case HttpURLConnection.HTTP_FORBIDDEN:
                             case HttpURLConnection.HTTP_UNAUTHORIZED:
                                 reportIssue(HealthIssue.SERVER_AUTHENTICATION_ISSUE);
+                                mHandler.getLooper().quit();
```
capnfabs: Nice. It might be better to call `stop()` or `stopImpl()` instead of quitting the Looper directly though.

dancunningham:
It will still show the error message right? And if you change the password or reapply settings will it resume checking?

llvasconcellos:
@dancunningham yes!

app/build.gradle
```diff
@@ -161,6 +161,8 @@ def requireWifi = project.requireWifi ?: 'false';
 // database unencrypted so that it can be easily inspected using Stetho.
 def encryptionPassword = project.encryptionPassword ?: '';
 
+def syncAccountName = "sync";
```
capnfabs, line 164: Is there a reason for defining this in the `build.gradle` instead of in the code where it was defined before?

capnfabs:
Thanks for looking into this Leonardo!

As I mentioned to you in chat, I don't think the approach of disabling the sync account and re-enabling it again later is a good idea. Here's why:
- We'll still make enough requests that it would be possible to generate the auth errors. Other sources that generate requests are:
  - User syncing, which can happen independently of the actual sync process
  - Analytics / server-side logging, which happens on some button clicks and on every activity change
  
  Additionally, all health checks are restarted when a user switches to any activity except for the settings activity, including the `BuendiaApiHealthCheck`.
- The logic for whether requests should be started or stopped is now split across multiple places:
  - `BuendiaApiHealthCheck`
  - `SyncAdapter`, which checks for an `AuthFailureError` as the cause of the exception a few layers deep.
  - The settings code, which is responsible for restarting everything when settings change just in case it works.

Instead, I would advise overriding `BuendiaApiHealthCheck#isApiUnavailable()` to return a value depending on auth failure, which influences `HealthMonitor#isApiUnavailable()`. Then, you can make the other request sources check that value before querying the API. This means all authorisation checking logic is centralised in that single health check, which reduces the risk of bugs, and makes it easier to modify in the future.

llvasconcellos:
Ok @capnfabs I'll look at this after this sprint. I'll leave the PR open though.

llvasconcellos:
I'm still leaving it open...

llvasconcellos:
I went for a different approach following @capnfabs advice.
https://github.com/projectbuendia/client/pull/275

PR #151, opened 2015-11-18
Don't export our Odk Content Providers anymore.

capnfabs:
The ODK content providers are only used internally, not by any other app. As such, they don't need to be exported.


llvasconcellos:
Great. Good to go!

PR #152, opened 2015-11-18
Better client-side logging for Observation sync.

capnfabs:


llvasconcellos:
Nothing to add. Good to go!

PR #154, opened 2015-11-20
Incremental sync for Orders.

capnfabs:


app/src/main/java/org/projectbuendia/client/json/JsonOrdersResponse.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * JSON representation of a set of patients returned by the server.
+ * <p>
+ * TODO: Generify this class into e.g. an IncrementalFetchResponse<JsonEncounter>
+ */
+public class JsonOrdersResponse {
+    public JsonOrder[] results;
+    // TODO(capnfabs): Rename this to syncToken.
+    /** In ISO 8601 date format. */
+    public String snapshotTime;
```
llvasconcellos, line 12: Jax-RS @produces and @consumes are so much more fun. You would use DateTime here.

llvasconcellos:
Looks good to me.

app/src/main/java/org/projectbuendia/client/json/JsonOrdersResponse.java
capnfabs, line 12: Gson supports parsing things into dates (see `org.projectbuendia.client.json.Serializers`), but we shouldn't actually be treating this as a date - that's a legacy thing. It's an opaque string that we're just supposed to send to the server. I'll rename all of these from `snapshotTime` to `syncToken` later this sprint, and then we'll remove the reference to them being dates.

PR #158, opened 2015-11-26
Changing require wifi default to true.

### LINKED ISSUE #149: Snackbar no longer picking up wi-fi status
It looks like Snackbar behaviour has changed. When I switch off wi-fi, it used to say “Wi-fi disconnected” with a “Connect” button. Now it goes straight to “Server unreachable” with a “More info” button. The message should depend on the detected situation.

llvasconcellos:
This fixes #149

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -154,7 +154,7 @@ def openmrsUser = project.openmrsUser ?: openmrsUserDefault
 def openmrsPassword = project.openmrsPassword ?: openmrsPasswordDefault
 def packageServerRootUrl = project.server ? 'http://' + project.server + ':9001'
     : project.packageServerRootUrl ?: packageServerRootUrlDefault
-def requireWifi = project.requireWifi ?: 'false';
+def requireWifi = project.requireWifi ?: 'true';
 
 // This is the password used to encrypt the SQLite database on the tablet.
 // In development builds, the password is an empty string, which leaves the


capnfabs:
Looks good to me.

FYI on Release builds (which we're not using at the moment) and debug builds built by jenkins, Wifi does default to true - what you're observing here only applies to local builds.

dancunningham:
@capnfabs I just installed from the client-master build and "Require wifi connection" is not set to true.

dancunningham:
Oh, maybe because I didn't erase data.

But not sure why I would ever have set that to false? Was false the default previously?

llvasconcellos:
@dancunningham That option didn't exist before. The app required wifi as a default behavior.
After I talked to @zestyping he added this option and default it to false.
I don't know about jenkins build though.

llvasconcellos:
We also could document the Require WIFI on the wiki.

zestyping:
The original motivation for making it false by default in dev builds was
that we run dev builds in the emulator so often.  In the emulator there is
no wifi, so when "Require wifi connection" was true, the snackbar would
always be complaining about wifi, which prevented us from seeing or testing
any of the other snackbar messages reliably — for example, app updates were
inaccessible.

capnfabs:
@dancunningham if the settings aren't found, the defaults are written on first boot. So even with this change, the old default setting would persist if you hadn't cleared data.

In my opinion, this isn't ideal and is probably something we should change.

capnfabs:
I'm going to close this change - as stated above:

> Release builds (which we're not using at the moment) and debug builds built by jenkins, Wifi does default to true - what you're observing here only applies to local builds.

I think that's good enough.

PR #161, opened 2015-11-30
Cleaning up OdkActivityLauncher class

viniciusboson:
I've cleaned up the `OdkActivityLauncher` class to be more readable and maintainable. 

There were many methods too long and they were doing to many things that it was necessary to debug them paying very close attention to understand what they were supposed to do.

Theses changes don't have any intention to cause side-effects. This is just a janitor PR which will be useful for future changes.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/exception/ValidationException.java
+++ b/app/src/main/java/org/projectbuendia/client/exception/ValidationException.java
@@ -1,0 +1,35 @@
+// Copyright 2015 The Project Buendia Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License.  You may obtain a copy
+// of the License at: http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distrib-
+// uted under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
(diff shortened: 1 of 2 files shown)


capnfabs:
Awesome, thanks for doing this! That class really needed some work. I've left some comments on the individual commits but after that it's good to go :+1:

capnfabs:
This is good to go when you're ready :+1:

viniciusboson:
@capnfabs, thx for you opinion ...
I've chosen a try-catch approach just to centralize both event error handling and method exit point.
I've removed the validations methods too because their declaration + comments are longer than themselves ... 
I've pushed an prior commit with trigging the events each time too ...
What do you think about it ?

capnfabs:
Hey @viniciusboson, nice work! I like the try/catch mechanism. Two notes:
- Just because the method declaration and comments are longer, doesn't mean it's not worth refactoring them out :) It's ok so long as the code is easier to understand, even if the actual file is longer.
- If it's important that the exceptions are caught or exceptions, throw checked exceptions rather than unchecked exceptions. `IllegalStateException` is an unchecked exception.

viniciusboson:
Hey @capnfabs. I've done the changes that you suggested. I created a checked ValidationException class and rolled-back to keep validations inside each method.

capnfabs:
Thanks Vinicius! This is good to go, feel free to merge when you're ready.

PR #162, opened 2015-12-01
Add support for paginated observation sync.

capnfabs:
- Use logic similar to PatientSyncPhaseRunnable in ObservationSyncPhaseRunnable. These will be
  unified into a generic, shared-code system later.
- Fix the source of slowness in Observation sync by executing the DB ops in a single bulk
  transaction.


llvasconcellos:
Fine by me. Good to go!

PR #166, opened 2015-12-02
Reduce form download logging.

capnfabs:
Previously, each form downloaded would be dumped to logcat. This is super noisy and unnecessary.
This change logs the UUID of the downloaded form instead.


llvasconcellos:
Fine by me. Good to go.

PR #167, opened 2015-12-02
Genercise client-side incremental sync logic.

capnfabs:
ObservationsSyncPhaseRunnable and PatientsSyncPhaseRunnable both had lots of common code, and there
were small inconsistencies between them that went unnoticed. This change centralises the common
logic into a base `IncrementalSyncPhaseRunnable`, which will also be used for Orders in a future
change.

Some notes about the design:
- I'm normally opposed to inheritance, but in this situation, it's syntactically a lot more
  convenient, and the expected relationship between the superclass and the subclass is well-defined.
- This is the reason why I've marked the `#sync` method as final and gone for "beforeSyncStarted"
  and "afterSyncFinished" callbacks instead of getting subclasses to override the `#sync` method -
  it's much closer to a composition-style design in this manner.


app/src/main/java/org/projectbuendia/client/json/IncrementalSyncResponse.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public String syncToken;
+
+    /**
+     * {@code true} if there is more data that remains unfetched. In this case, a subsequent request
+     * to the same endpoint that supplies {@link #syncToken} from this response will return more
+     * data that the client hasn't previously fetched.
+     */
+    public boolean more;
+}
```
llvasconcellos, line 36: :+1:

app/src/main/java/org/projectbuendia/client/sync/controllers/IncrementalSyncPhaseRunnable.java
```diff
@@ -0,0 +1,182 @@
...
+                connectionDetails.addAuthHeader(new HashMap<String, String>()),
+                successListener,
+                wrapErrorListener(errorListener));
+        Serializers.registerTo(request.getGson());
+        request.setRetryPolicy(
+                new DefaultRetryPolicy(Common.REQUEST_TIMEOUT_MS_MEDIUM, 1, 1f));
+        connectionDetails.getVolley().addToRequestQueue(request);
+    }
+
+    private static class IncrementalSyncResponseType implements ParameterizedType {
```
llvasconcellos, line 159: Why did you make this class static?

llvasconcellos:
Great design. I like it. Good to go.

app/src/main/java/org/projectbuendia/client/sync/controllers/IncrementalSyncPhaseRunnable.java
capnfabs, line 159: It's a static nested class because it doesn't know anything about the outer class it's contained in. See https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html

PR #171, opened 2015-12-04
Removing old snackbar.

llvasconcellos:
Migration of error messages to new snackbar and removing the old one.
The messages don't pile up anymore!

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
@@ -349,49 +349,31 @@ public abstract class BaseActivity extends FragmentActivity {
 
     protected class UpdateNotificationUi implements UpdateNotificationController.Ui {
 
-        final View mStatusView;
-        final TextView mUpdateMessage;
-        final TextView mUpdateAction;
-
-        public UpdateNotificationUi() {
-            mStatusView = getLayoutInflater().inflate(R.layout.view_status_bar_default, null);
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
```diff
@@ -132,6 +129,12 @@ public int snackBar(String message, String actionMessage, View.OnClickListener l
         return snackBar.message(message, actionMessage, listener, 999);
     }
 
+    public int snackBar(int resourceIdMessage, int resourceIdActionMessage, View.OnClickListener
```
capnfabs: Use the `@StringRes` annotation for this and Android Studio will warn you if you accidentally pass the wrong type of value in here :)

capnfabs:
I think I would do it at the snackbar level similarly to how [Android does notifications](http://developer.android.com/guide/topics/ui/notifiers/notifications.html#Updating). For notifications, you have to pass a unique numeric ID to the notification function, which defines whether it overwrites an existing notification or not. For snackbar, this means you'd have to re-architect some of the functions to accept an ID instead of generate an ID, but I think this is a pretty clean solution.

dancunningham:
Not allowing duplicates sounds good to me, and if a notification is triggered a second time, it should bump up to the top of the list within its priority level.

capnfabs:
@dancunningham just have to be careful then to assign different priorities to all the health check messages, lest they keep swapping positions arbitrarily.

Issue #125, opened 2015-11-05
PatientChartActivity drops temporary observations prematurely.


capnfabs:
See `PatientChartActivityTest` for reference.

Both of our tests that create multiple observations for a single patient in quick succession are flaky, and only pass about half of the time. The cause of the test failures appears to be that the chart fails to update after observations are made sometimes.

The failure mode for `testCombinesNonOverlappingObservationsForSameEncounter` is that sometimes, all of the observations from a single encounter are missing.

For `testEncounter_latestEncounterIsAlwaysShown`, sometimes the observations update to the new values, and sometimes they don't.

llvasconcellos:
testEncounter_latestEncounterIsAlwaysShown test actually found a bug. I problem goes deeper and is documented here:
https://github.com/projectbuendia/client/issues/85
https://github.com/projectbuendia/client/pull/90

That's why PR #90 isn't merged yet. It should fix this behavior but it doesn't.

capnfabs:
Thanks for the context! I suspect there's multiple issues here:
- The incremental sync issue that @zestyping sent PR #90 for
- Whatever is causing that to not fix the problem that you're observing on the same PR.

capnfabs:
OK. I _think_ the problem is:
- Sync for Observations starts. Results are fetched, but JSON isn't parsed (yet).
- Simultaneously, the Xforms request is completed, and the success callback which adds the temporary records to the database is completed on the main thread.
- The JSON is parsed for Observations sync. The new records are added to the database, and the temporary records are deleted.

Thus, the temporary records are deleted without actually being used. This is why we see the flicker from the correct value back to the old value.

I'm not sure if this is the only cause but it's a start.

capnfabs:
Another possible cause:
- the `encounter_millis` value we insert for temporary observations currently comes from the ODK form, which is truncated down to the minute. This means that the code that computes the latest observations will return a synced observation over a locally inserted observation every time.

capnfabs:
The final possible cause:
- The code that requests periodic syncs runs the first one as soon as the chart is loaded. This drastically increases the probability that the first cause occurs.

capnfabs:
From the Xforms code on the server:

<img width="550" alt="screen shot 2016-01-27 at 7 18 06 pm" src="https://cloud.githubusercontent.com/assets/1336508/12625242/d5a5fe18-c52a-11e5-933b-3dab61003e8a.png">

That's a good place to start.

llvasconcellos:
[...]

PR #172, opened 2015-12-04
[2016] Submit forms offline

viniciusboson:
This feature allows users to save observation forms offline and synchronize them as soon as server connection gets available.
The app will show an alert message in Snackbar whenever it detects that there are unsent forms.
To test this feature, i've disabled the wifi to not be able to access dev.projectbuendia.org or i've simulated a local server always throwing exceptions.
There are three concurrent situations that the application will try to resent the forms:
1) On observation sync process, just before get all fresh observations
2) Prior to sent a new observation requested by the user
3) When unsent form health check detects that there are still pending forms to be submitted.

screenshot_2016-01-22-08-10-47

For further informations, please refer to its [documentation](http://tinyurl.com/px68kuw)


capnfabs:
Hey Vinicius, I've been away. Where is this up to? Is it ready to be merged?

capnfabs:
Hey Vinicius, this PR was messier than I thought. Could you merge `dev` into this branch? Parts of the code at HEAD are in your PR, and it's difficult to tell them apart.

app/src/main/java/org/projectbuendia/client/models/UnsyncForm.java
```diff
@@ -0,0 +1,86 @@
...
+import android.content.ContentValues;
+import android.database.Cursor;
+
+import org.projectbuendia.client.providers.Contracts;
+import org.projectbuendia.client.utils.Utils;
+
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public final class UnsyncForm  {
```
capnfabs: Can we use a name like "UnsentForm" or "PendingForm" instead? "UnsyncForm" isn't clear which direction the form is going in (client to server or server to client).

app/src/main/java/org/projectbuendia/client/models/UnsyncForm.java
```diff
@@ -0,0 +1,86 @@
...
+    public final String uuid;
+    public final String patientUuid;
+    public final String xml;
```
capnfabs: Suggest naming this field `formContents`

app/src/main/java/org/projectbuendia/client/models/UnsyncForm.java
```diff
@@ -0,0 +1,86 @@
...
+    /** Puts this object's fields in a {@link ContentValues} object for insertion into a database. */
+    public ContentValues toContentValues() {
+        ContentValues cv = new ContentValues();
+        cv.put(Contracts.UnsyncForms.UUID, uuid);
+        cv.put(Contracts.UnsyncForms.PATIENT_UUID, patientUuid);
+        cv.put(Contracts.UnsyncForms.XML, xml);
+        return cv;
+    }
+
+    public static final class Builder {
```
capnfabs: This class doesn't have many members, so I would advise using a constructor that allows you to set the args instead of a builder. e.g.

``` java
public UnsyncForm(String formUuid, String patientUuid, String xml) { //...
```

app/src/main/java/org/projectbuendia/client/providers/Contracts.java
```diff
@@ -36,7 +36,8 @@
         ORDERS("orders"),
         PATIENTS("patients"),
         USERS("users"),
-        SYNC_TOKENS("sync_tokens");
+        SYNC_TOKENS("sync_tokens"),
+        UNSYNC_TOKENS("unsync_forms");
```
capnfabs: This should be consistent with the data type you've created and the table name.

app/src/main/java/org/projectbuendia/client/sync/Database.java
```diff
@@ -160,6 +160,12 @@
         SCHEMAS.put(Table.SYNC_TOKENS, ""
             + "table_name TEXT PRIMARY KEY NOT NULL,"
             + "sync_token TEXT NOT NULL");
+
+        SCHEMAS.put(Table.UNSYNC_TOKENS, ""
+            + "uuid TEXT PRIMARY KEY NOT NULL,"
+            + "patient_uuid TEXT,"
+            + "xml");
```
capnfabs: It looks like you're missing a column type for this line?

app/src/main/java/org/projectbuendia/client/sync/Database.java
```diff
@@ -160,6 +160,12 @@
         SCHEMAS.put(Table.SYNC_TOKENS, ""
             + "table_name TEXT PRIMARY KEY NOT NULL,"
             + "sync_token TEXT NOT NULL");
+
+        SCHEMAS.put(Table.UNSYNC_TOKENS, ""
+            + "uuid TEXT PRIMARY KEY NOT NULL,"
+            + "patient_uuid TEXT,"
```
capnfabs: Should this be `TEXT NOT NULL` ?

app/src/main/java/org/projectbuendia/client/sync/controllers/IncrementalSyncPhaseRunnable.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public final void sync(ContentResolver contentResolver, SyncResult syncResult,
+        ContentProviderClient providerClient) throws Throwable {
+
+        boolean okToProceedSync = beforeSyncStarted(contentResolver, syncResult, providerClient);
+
+        if(!okToProceedSync) {
+            LOG.w("Skipping synchronization for %s", this.getClass().getName());
```
capnfabs: Use `getClass().getSimpleName()` instead for this, `getName()` is the fully-qualified verbose name.

app/src/main/java/org/projectbuendia/client/sync/controllers/IncrementalSyncPhaseRunnable.java
```diff
@@ -0,0 +1,190 @@
...
+        this.resourceType = resourceType;
+        this.dbTable = dbTable;
+        this.clazz = clazz;
+    }
+
+    @Override
+    public final void sync(ContentResolver contentResolver, SyncResult syncResult,
+        ContentProviderClient providerClient) throws Throwable {
+
+        boolean okToProceedSync = beforeSyncStarted(contentResolver, syncResult, providerClient);
```
capnfabs, line 88: This is a really elegant design. Nice work Vinicius!

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartController.java
```diff
@@ -269,21 +269,21 @@ public void suspend() {
...
-            ? "form_discard_pressed" : "form_save_pressed";
-        Utils.logUserAction(action,
-            "form", request.formUuid,
-            "patient_uuid", request.patientUuid);
-        mOdkResultSender.sendOdkResultToServer(request.patientUuid, resultCode, data);
-        mUi.showFormSubmissionDialog(shouldShowSubmissionDialog);
+        final boolean isSubmissionCanceled = (resultCode == Activity.RESULT_CANCELED);
+        Utils.logUserAction(isSubmissionCanceled ? "form_discard_pressed" : "form_save_pressed",
+            "form", request.formUuid, "patient_uuid", request.patientUuid);
+        if(isSubmissionCanceled) return;
```
capnfabs, line 301: Now that you're checking this here, you can remove the parameter and check from `OdkResultSender#sendOdkResultToServer`. Thanks for doing this, it's much more readable.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -293,102 +279,146 @@ private static OpenMrsXformIndexEntry findUuid(
...
-                    new SubmitXformFailedEvent(SubmitXformFailedEvent.Reason.CLIENT_ERROR));
-                return;
+
+            final Long formIdToDelete = getIdToDeleteAfterUpload(context, uri);
+            if(!validateIdToDeleteAfterUpload(formIdToDelete, uri)) {
+                throw new ValidationException("No id to delete for after upload: " + uri);
             }
-            final long idToDelete = instanceCursor.getLong(columnIndex);
 
             // Temporary code for messing about with xform instance, reading values.
```
capnfabs: Is this still "temporary code"?

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -293,102 +279,146 @@ private static OpenMrsXformIndexEntry findUuid(
...
+     * Validates the xml. Returns {@code true} if it is valid. Otherwise, returns {@code false}
+     */
+    private static boolean validateXml(String xml) {
+        return xml != null;
+    }
+
+    private static void deleteLocalFormInstances(Long formIdToDelete) {
+        //Code largely copied from InstanceUploaderTask to delete on upload
+        DeleteInstancesTask dit = new DeleteInstancesTask();
+        dit.setContentResolver(
```
capnfabs: It's probably better to do this in the constructor than in a separate setter.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -396,118 +426,277 @@ public static void sendOdkResultToServer(
...
                     }
+                }, new Response.ErrorListener() {
+                    @Override public void onErrorResponse(VolleyError error) {
+                        //Just log it and flag as not synchronized, this form is already persisted.
+                        LOG.e(error, format("Error resubmitting %s form to server ", unsyncForm.uuid));
+                        hasUnsubmittedForms[0] = true;
+                    }
+                });
+        }
+        return !hasUnsubmittedForms[0];
```
capnfabs: I don't think this will always return the right value - you need to block until one of the response listeners has been called to know for sure whether submission worked or not.

capnfabs:
Please merge https://github.com/projectbuendia/client/pull/161 into `dev` and then `dev` into this branch, which should make what's changed in the pull request much clearer.

This is looking really good! I left a few comments but after those have been addressed OK to merge. :+1: 

Thanks!

app/src/main/java/org/projectbuendia/client/providers/Contracts.java
zestyping: What's the UX story for this?  Sorry if I've missed a design discussion
about this.

As I last remember, we weren't going to do offline submission because there
were too many unsolved UX design issues related to users understanding when
their changes had actually been saved, and how to handle edit conflicts.
Since we thought that not knowing which orders had been administered could
lead to life-threatening mistakes, we had decided it was too risky.

Were these design issues ever addressed?  If they weren't, I would be very
afraid to land this feature.

capnfabs: The plan UX-wise was to add a permanent snackbar thingy with the number of pending observations and a "retry" button. Edit conflicts aren't a problem for observations because you can't edit them :)

Also I don't think that executing orders uses this mechanism at the moment - it's just forms as of this change.

cc @dancunningham

app/src/main/java/org/projectbuendia/client/models/UnsyncForm.java
viniciusboson: Sure, @capnfabs. Agreed with you. I'll refactor it to 'UnsentForm'.

app/src/main/java/org/projectbuendia/client/models/UnsyncForm.java
viniciusboson: Great, thx ...

viniciusboson:
Thx for your review, @capnfabs ... I told you that i've messed it with all those merges ..  :-/
I'll clean it up, implement your suggestions and let you know.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartController.java
viniciusboson, line 301: @capnfabs, `OdkResultSender#sendOdkResultToServer`can still not submit the form, when it gets a validation exception. In this case, the form submition dialog would pop up indefinetely. So the second validation `isSubmittingForm = mOdkResultSender.sendOdkResultToServer(...)` is to prevent these scenario. 
If you take a look at `sendOdkResultToServer` implementation, you'll see that i returns `true` wether or not it is submitted properly, and returns false in case of validation failure. 
This behavior bothers me a bit, but i would not like to accouple `sendOdkResultToServer` with `mUi.showFormSubmissionDialog(isSubmittingForm);`, puting it inside the method. 
What do you think about that ?

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson: Good catch, @capnfabs ... Could you give me a hint in how to block until one of these events are completed ?

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartController.java
capnfabs, line 301: I'm not sure we're talking about the same thing :)

I meant that now that you're checking whether submission was cancelled here (`if(isSubmissionCanceled) return;`), you can remove the identical check in OdkResultSender#sendOdkResultToServer, on the line `if(isActivityCanceled(resultCode, data)) return false;`.

Does that make sense?

viniciusboson, line 301: Yeah, we weren't ... lol .. sorry about that ...
But, if possible, could you check that comment too ? I would like to know your opinion about that.

Getting back to the point you've commented, i had left that check inside `OdkResultSender.sendOdkResultToServer` too because i thought that this method could be called by other controllers. But i agree with you. It is useless now. I'll delete it. Thx

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
capnfabs: Sure. Take a look at this Gist, I've written a couple of examples for you: https://gist.github.com/capnfabs/d9e8dac58d23ac89c838

Normally, if you just had one thing you were submitting, you would define a `Future`, set the Future in the callbacks, and then wait on the future before returning. The idea with `Futures` is that it represents a value that will be available in the future, but isn't yet. I've demonstrated this in `WaitForSingleForm.java`.

Because we've got multiple forms instead, we can wait for all of them to complete by using a `CountDownLatch`. I've included an example for this in the gist, the example file is called `OdkActivityLauncher.java`. `CountDownLatch`es block until they've counted down to 0, so in this example, we create a CountDownLatch with a size equal to the number of requests we're going to send, and then when each request either completes or fails, we count down. In the `resendFormsToServerBlocking` method, we wait until the CountDownLatch has counted down completely, and then return the value of `hasUnsubmittedForms`, which we know will be up to date.

Does that make sense?

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartController.java
capnfabs, line 301: Sure. As for your other comment, I'm ok with `sendOdkResultToServer` returning `true` so long as it attempted to send. The other way to do it would be to split it up into a `validate` method and a `send` method, but I think the current approach is simple enough to understand that I don't mind.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
viniciusboson: It makes totally sense .. Both code and comment are clear enough. Thanks a lot, @capnfabs.

app/src/main/java/org/projectbuendia/client/providers/Contracts.java
zestyping: Ok, I think that's okay as long as:
1. Order execution is synchronous-only
2. Pending submissions are retried automatically when you submit the form

I think #1 is important enough that it deserves a comment explaining the design decision.  The rationale for #2 is that it would probably be confusing if a user submits some information (A) and it gets queued, and then later the user submits some more information (B) and B gets saved but A is still not saved.

@dancunningham Any thoughts on the UX for this?  Is there a document or wiki page describing the technical design or UX design for this feature?

capnfabs: (2) is a good idea. Additionally, we also try to send any pending data with every sync, which, while in the patient chart, is every minute.

Regarding (1), I think you're right - we need to explicitly call out and explain this.

We don't have any design notes about this except for what we've got in [Pivotal Tracker](https://www.pivotaltracker.com/story/show/107200344).

dancunningham: Agree on (1). @viniciusboson can you add a comment noting that Order Execution is synchronous-only and the reason why (to prevent double administering of treatments)

On (2) I also agree it would be especially confusing if (A) and (B) were both on the same patient. If we're going to make it attempt submit queued submissions when submitting a form, we might as well send them all.

@zestyping: there is technical design documentation in this document: https://docs.google.com/document/d/1ah6Ezxc7Fhh3CcmoWTPUhbkYCCkiOnPbGHJ5038u7dc/edit

viniciusboson: @dancunningham, @capnfabs @zestyping  what do you all mean by 'Order execution is synchronous-only' ? When app tries to submit the order, if it's ok, there is nothing to do. If not, it goes to an unsent queue.
This is already synchronized, right ? It will need to wait for server response anyways. Does it make sense?

About (2), that's a really good point, @zestyping. I was doing as @capnfabs said, trying to send forms before pulling new forms. But we can add the behavior to when it tries to submit any form tom. Thx for the suggestion

capnfabs:
@viniciusboson I'm not sure what happened to your comment, so I'll write back here.

When @zestyping says "Order execution is synchronous-only", he means that when orders are executed, they submit direct to the server, or fail visibly, so the user knows that the order execution wasn't recorded on the server. So there is no offline caching for order execution like there is for filling in observations.

This is important for Orders, because otherwise it could lead to patients getting multiple doses of the same drug.

As written, your change already does this, so it's all good! But it's probably important to note in a comment somewhere that we don't allow order execution to be cached offline and submitted later.

capnfabs:
To sum up the conversation that @viniciusboson and I had today:
- There's currently a deadlock in form submission in this PR because
  1. We block on the main thread until all the Volley form network requests complete == all unsent forms are sent
  2. The form network request callbacks are run on the main thread, but they never run because of (i).
- The solution (and a very important change anyway for UI responsiveness) is to take the blocking call off the main thread and onto a background thread.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -451,6 +450,17 @@ public void updatePatientLocationUi(LocationTree locationTree, Patient patient)
...
+            PatientChartActivity.this.snackBar(errorMessageResource,
+                R.string.submit_xform_resubmit,
+                new View.OnClickListener() {
+                    @Override public void onClick(View view) {
+                        OdkActivityLauncher.submitUnsetFormsToServer(
+                            PatientChartActivity.this.getContentResolver());
+                    }
+                }, 995, false); //TODO: Define proper priority
+        }
+
```
viniciusboson: @dancunningham: What should this priority be ? Should it be dismissible ?

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartController.java
```diff
@@ -694,7 +695,7 @@ public void onEventMainThread(SubmitXformFailedEvent event) {
                 default:
                     errorMessageResource = R.string.submit_xform_failed_unknown_reason;
             }
-            mUi.showError(errorMessageResource);
+            mUi.showFormSubmissionError(errorMessageResource);
```
viniciusboson: @dancunningham: Do you think these messages should be maintained ? I've created a new message entry to alert the user that the form was not sent and it will be send later.

app/src/main/res/values/strings.xml
```diff
@@ -222,6 +222,8 @@
   <string name="submit_xform_failed_unknown_reason">Something went wrong submitting the form.</string>
   <string name="submit_xform_failed_server_auth">Error submitting form: could not authenticate with server.</string>
   <string name="submit_xform_failed_server_timeout">Timed out while sending form to the server.</string>
+  <string name="submit_xform_failed_resend_later">Error submitting form to the server: It will be saved to be resent later.</string> <!-- TODO: Use a better msg -->
```
viniciusboson: @dancunningham: Could you help me with a more meaningful message ?

capnfabs:
Hey Vinicius, regarding how to take the blocking form submit off the main thread, you can do this by chaining asynchronous calls together. So, as an example:
- In `sendOdkResultToServer`, you can call `submitUnsetFormsToServer` on a background thread using an AsyncTask (or similar).
- Once `submitUnsetFormsToServer` finishes, continue on with the existing logic that's already in `sendOdkResultToServer`.
- The code that returns form submission status back to the UI is already asynchronous and runs on the event bus, so the form dialog should dismiss itself automatically.

Hope that helps :)

dancunningham:
I'll test this today!

llvasconcellos:
Amazing work @viniciusboson!

dancunningham:
Just tested this, first in the situation of being connected to the server, and I saw the following:
1. Opened anthropometrics form and filled it in
2. Data appeared in chart
3. Then after a few seconds it disappeared
4. Then after about 30s it appeared again

I checked on the server via web interface and it had been saved correctly.

(3) shouldn't have happened.

viniciusboson:
Thx, @llvasconcellos :)

viniciusboson:
@dancunningham ,

I've noticed that too. I suspect that (3) happens due observation sync process lag.
Once it detects that there is no more unsent forms, it cleans all cached forms but this one that it has been removed it is not fetched yet. Then, on the next phase, it fetches properly.
Does it make sense, @capnfabs ?

capnfabs:
When a form gets submitted, the observations get stored locally before the next sync so that the chart looks right. @dancunningham, was this a one-off? We've got a test that's been flaky for ages because of this #125.

@viniciusboson, code quality looks good, I'm happy with this now. Thanks so much! Feel free to merge once Dan confirms whether the bug he's observing is a repeat occurrence or a one-off.

viniciusboson:
@capnfabs, I can reproduce this issue every time i'm creating an observation too.
One possible solution that i'm thinking about is the method `ObservationsSyncPhaseRunnable#afterSyncFinished` just delete the temporary observations with `encounter_millis` less than the sync_token's millis. Sounds reasonable for you, this approach?

capnfabs:
@viniciusboson I'd avoid that - the sync token is supposed to be opaque to the client.

Deletion of temporary observations only occurs if the sync completes successfully, and the sync only completes successfully if all unsent forms have previously been added to the server. So it shouldn't be possible for sync to miss the forms that haven't been submitted, right?

My guess is that the offline forms aren't being submitted synchronously as part of the sync process any more. I think we should fix that instead of adding a workaround to the observations syncing code.

dancunningham:
@capnfabs @viniciusboson I'm not seeing it happen every time, just sometimes, so let's continue to track that in #125 and not let that block merging this in.

Have a little more testing to do on this today before we merge!

dancunningham:
Please bump up the priority of the snackbar message so that it appears above "Serveur inaccessible".

If disconnected, the server inaccessible message will already be there, so you're only going to notice the "pending submission" message if it's higher priority.

viniciusboson:
@dancunningham: I've raised the submission offline alert. Have you found any error else?
@capnfabs: I got you point. I'm trying to fix this race condition. I coundn't fix it yet but i'm working on that.

PR #178, opened 2015-12-07
Last snackbar fixes

llvasconcellos:


capnfabs:
Nice work Leonardo! Ok to merge when you're ready :+1:

llvasconcellos:
There's no reason. I just discovered View.INVISIBLE and decided it's more suitable for layout needs. As explained here:
http://stackoverflow.com/questions/11556607/android-difference-between-invisible-and-gone

capnfabs:
Oh, cool. Yeah, that's spot on - it's much better for this situation because of the way it affects layout.

PR #181, opened 2015-12-09
Remember chart scroll position

llvasconcellos:
Remembering scroll position and applying it after the chart finished loading.

**Obs: The previous behavior is maintained. (The chart is being scrolled to the right after loading).**
So, for now, only the Vertical position is restored because the horizontal is being updated elsewhere (javascript).
@dancunningham Do we need to remember the horizontal scroll position as well?

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -184,8 +184,13 @@
 
   $(document).ready(function() {
     $('#grid').freezeHeader({top: true, left: true});
-    $('#grid-scroller').scrollLeft(1e6);
   });
+
+  $( window ).unload(function() {
+      console.log("Window => v=" + $(window).scrollTop() + " - h=" + $(window).scrollLeft());
+      console.log("grid-scroller => v=" + $('#grid-scroller').scrollTop() + " - h=" + $('#grid-scroller').scrollLeft());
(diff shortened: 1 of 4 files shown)


dancunningham:
I've tested this. It is acceptable for the moment, and better than before.

It works pretty well for updating the Treatment Plan with the jump up and down pretty snappy (within 1 second).

Eventually I'd like this to be a lot slicker, without any jumping up and down, which I believe would mean updating the DOM directly via Javascript. I understand that's a significant change.

capnfabs:
FYI @dancunningham, updating the DOM directly via JS is something we've talked about in the past for code health reasons too - the whole view rendering system would be more cleanly separated switching to some kind of JS framework for UI rendering, as opposed to the Java / Pebble template / JS mess that we've got at the moment. As a result, it would also arguably reduce the barrier to entry for developers on the project, because they could still do chart work without having to understand the rest of the stack.

llvasconcellos:
@capnfabs I agree!

zestyping:
Just a note of concern — although migrating to JS-only rendering does seem
feasible, it may be more difficult that it seems at first glance.  There
are some good reasons we have the design we currently have.  We have a
pretty flexible templated formatting framework right now that powers much
of the customization of the chart view.  To do this in JS would require
migrating that logic to JS as well as converting the Obs data types to a JS
representation.

PR #182, opened 2015-12-09
Show problem resolved

llvasconcellos:
Show a snackbar message with troubleshooting problem resolved.

@dancunningham I wasn't feeling creative with the messages. Please review them. We also need to translate them to French.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/diagnostics/Troubleshooter.java
+++ b/app/src/main/java/org/projectbuendia/client/diagnostics/Troubleshooter.java
@@ -48,14 +48,14 @@ public class Troubleshooter {
     }
 
     /**
-     * Returns true iff the given issue is current active.
+     * Returns true if the given issue is current active.
      * @param issue {@link HealthIssue} to check for
      */
     public boolean hasIssue(HealthIssue issue) {
         return mActiveIssues.contains(issue);
     }
 
-    /** Returns true iff no active issues exist. */
+    /** Returns true if no active issues exist. */
     public boolean isHealthy() {
         return mActiveIssues.isEmpty();
     }
@@ -108,10 +108,10 @@ public class Troubleshooter {
         // TODO: Consider scheduling this for ~100 milliseconds in the future so as to
         // prevent multiple troubleshooting events from firing for issues resulting from the same
         // root cause.
-        postTroubleshootingEvents();
+        postTroubleshootingEvents(null);
     }
 
-    private void postTroubleshootingEvents() {
+    private void postTroubleshootingEvents(HealthIssue solvedIssue) {
         synchronized (mTroubleshootingLock) {
             ImmutableSet.Builder<TroubleshootingAction> actionsBuilder = ImmutableSet.builder();
 
@@ -129,7 +129,7 @@ public class Troubleshooter {
             }
 
             mLastTroubleshootingActionsChangedEvent =
-                new TroubleshootingActionsChangedEvent(actions);
+                new TroubleshootingActionsChangedEvent(actions, solvedIssue);
             mEventBus.postSticky(mLastTroubleshootingActionsChangedEvent);
         }
     }
@@ -158,6 +158,6 @@ public class Troubleshooter {
         // TODO: Consider scheduling this for ~100 milliseconds in the future so as to
         // prevent multiple troubleshooting events from firing for issues resulting from the same
         // root cause.
-        postTroubleshootingEvents();
+        postTroubleshootingEvents(healthIssue);
     }
 }
--- a/app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
@@ -337,27 +337,56 @@ public abstract class BaseActivity extends FragmentActivity {
     private void displayProblemSolvedMessage(HealthIssue solvedIssue) {
         Map<HealthIssue, Integer[]> troubleshootingMessages = new HashMap<>();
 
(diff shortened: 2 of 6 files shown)


dancunningham:
There's an issue with the resolution of troubleshoot_wifi_disabled.

Steps to reproduce:
1. Set "Require wifi connection" to true
2. Disable wifi - the snackbar message appears
3. Enable wifi (either via the button in the snackbar, or using Android UI)
4. Now "Wifi is disconnected" appears (correctly), but "Wifi is disabled" is still present.

In this case, since "Wifi is disconnected" appears, we probably don't want a "problem resolved" message for when wifi is enabled, but the "Wifi is disabled" message should just be dismissed.

dancunningham:
Another issue: if I dismiss a message (e.g. Wifi is disconnected) and then navigate to a different activity, the message immediately reappears. Is that what we want?

dancunningham:
I've documented here what I think the desired behaviour is, and the variances from this I see when testing the current implementation: https://docs.google.com/document/d/152cWIKNgiASVxqm548m6hYNwpvZyg9geR4_ESQ0TYZU/edit#heading=h.8cad856zlogq

Feel free to argue any points on the desired behaviour if you think there would be a better way to do some things!

llvasconcellos:
@dancunningham The message reappear because it was triggered again.

llvasconcellos:
About this comment: https://github.com/projectbuendia/client/pull/182#issuecomment-165126661
It's fixed now.

app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
```diff
@@ -205,137 +207,195 @@ public void snackBarDismiss(@StringRes int[] id) {
...
+    }
+
+    private void displayProblemSolvedMessage(HealthIssue solvedIssue) {
+        Map<HealthIssue, Integer[]> troubleshootingMessages = new HashMap<>();
+
+        // The troubleShootingMessages Map have the issue as the key and an integer array as
+        // value. This array is composed by:
+        // 0 - The message string id triggered by the issue.
+        // 1 - The new message string id (solved message).
+        // 2 - The timeout count for the solved message.
```
llvasconcellos, line 344: New timeout adjustment.

app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
```diff
@@ -205,137 +207,195 @@ public void snackBarDismiss(@StringRes int[] id) {
...
+        });
+        troubleshootingMessages.put(HealthIssue.SERVER_NOT_RESPONDING, new Integer[] {
+            R.string.troubleshoot_server_not_responding,
+            R.string.troubleshoot_server_not_responding_solved,
+            10
+        });
+        troubleshootingMessages.put(HealthIssue.PACKAGE_SERVER_HOST_UNREACHABLE, new Integer[] {
+            R.string.troubleshoot_package_server_unreachable,
+            R.string.troubleshoot_package_server_unreachable_solved,
+            5
```
llvasconcellos, line 383: Only message with 5 sec timeout.

app/src/main/java/org/projectbuendia/client/diagnostics/Troubleshooter.java
```diff
@@ -48,14 +48,14 @@ public Troubleshooter(EventBus eventBus) {
     }
 
     /**
-     * Returns true iff the given issue is current active.
+     * Returns true if the given issue is current active.
```
capnfabs, line 51: FYI I meant this as ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if), in case you run across it again. But it's also probably unnecessary for a boolean. :)

app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
```diff
@@ -205,137 +207,195 @@ public void snackBarDismiss(@StringRes int[] id) {
...
+                    break;
+                default:
+                    LOG.w("Troubleshooting action '%1$s' is unknown.", troubleshootingAction);
+                    return;
+            }
+        }
+    }
+
+    private void displayProblemSolvedMessage(HealthIssue solvedIssue) {
+        Map<HealthIssue, Integer[]> troubleshootingMessages = new HashMap<>();
```
capnfabs, line 338: It's better to use simple java objects for this kind of thing than `Integer[]` - using simple java objects makes the code more readable and less error prone.

So, the object would look something like this:

```
private static class TroubleshootingMessage {
   public final @StringRes int messageId;
   public final @StringRes int resolvedMessageId;
   public final int timeout;

   public TroubleshootingMessage(
       @StringRes int messageId, @StringRes int resolvedMessageId, int timeout) { ... }
}
```

Also, you can use `int[]` instead of `Integer[]`, because you're not using `null`s in the array.

app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
capnfabs, line 344: For duration variables, it's important to put the units in the description, e.g. millis or seconds.

app/src/main/java/org/projectbuendia/client/diagnostics/Troubleshooter.java
llvasconcellos, line 51: Sorry, I thought it was a typo.

app/src/main/java/org/projectbuendia/client/ui/BaseActivity.java
llvasconcellos, line 344: you are right.

PR #184, opened 2015-12-11
Ping's edit/delete orders

llvasconcellos:
Merged with dev.
It's working for editing all the order information and for deleting it.
There is a button for the function to Stop Order Now that isn't working See here: https://github.com/projectbuendia/client/commit/cdc52c34314ae604ec90b27c8d2f911994c36920
@dancunningham I'll try to make the button do it's magic.


capnfabs:
Well this was a doozy, but it's done.

Note: I've removed the "Stop Order Now" button for the time being as it wasn't wired up. We can add it back in later as a separate PR.

PR #190, opened 2015-12-18
Fixing Server Unreachable Message

### LINKED ISSUE #189: Server Unreachable Issue not being cleared after problem solved.
If I have two system health problems:

1 - Typo on the server address setting.
2 - Typo on the username or password setting.

The snackbar will display both messages:

1 - “Server unreachable”
2 - “Server username/password may be incorrect”

If I fix the typo on the server address setting the message number 1 should disappear but it's only disappearing when solving both problems.

llvasconcellos:
The problem was that the Health system only report problem solved if all the problems are solved.
It now reports server connectivity problem when they are solved.
This fixes #189


dancunningham:
Looks good, tested and it works as I expect

Issue #187, opened 2015-12-16
Initial Patient Sync not working


llvasconcellos:
When using a local server all the patients you create on the tablet are being synced to the server but if you clean the tablet storage the patients that you created before are not being synced back to the tablet.
The patients remain on the server with no problem.
I'm using this profile:
[nutrition-chad16-fr.txt](https://github.com/projectbuendia/client/files/64226/nutrition-chad16-fr.txt)

capnfabs:
Hey @llvasconcellos, could you provide some more information?
- When you say "clean the tablet storage" do you mean through the `App Settings --> Apply and clear local data` UI, or through the `Android Settings --> Buendia --> Clear Data` UI?
- Could you repro this 100% of the time?

I _think_ I've seen this before - if you clear data through the Android Settings but don't Force Stop first, any pending syncs keep running, which can put things in an indeterminate state.

llvasconcellos:
@capnfabs I need to retest it. I can't remember but I think I was using the App Settings -> Apply and clear local data.
I'll test it again and get back to you.

capnfabs:
Alright, thanks. There's a chance that "Apply and Clear local data" isn't clearing the sync status too, I need to check that.

capnfabs:
Ok, Apply and clear local data should work, it just drops all tables. I'm going to close this as not reproducible, if you spot it again please reopen and attach a bug report :)

llvasconcellos:
I used "Apply and clear local data" and now I have no patients.
[bugreport-2016-01-08-01-09-53.txt](https://github.com/projectbuendia/client/files/82067/bugreport-2016-01-08-01-09-53.txt)
![bugreport-2016-01-08-01-09-53](https://cloud.githubusercontent.com/assets/1710823/12189535/caef2a74-b5a5-11e5-8f57-29dcfacef5d8.png)

capnfabs:
Great, thanks for reopening!

capnfabs:
Logs show that the sync is running, but patients and orders both fetch zero records, and observations fetches 1 record. This could be consistent with a cached sync token?

capnfabs:
Ok, I managed to repro this, but then I just left the tablet for a bit and it sorted itself out. Did you observe that behaviour?

capnfabs:
I think it's a result of a cancellation firing during the initial sync.

llvasconcellos:
[...]

PR #196, opened 2016-01-08
Better logging for incremental sync.

capnfabs:
I'm hoping that this will help me track down the cause of #187.


llvasconcellos:
Nothing to add. Good to go.

PR #197, opened 2016-01-11
Multiple Charts

llvasconcellos:
To switch between charts swipe from left to right.


app/src/main/java/org/projectbuendia/client/models/Chart.java
```diff
@@ -16,11 +16,16 @@
 /** A chart definition. */
 public class Chart {
     public final String uuid;  // UUID of the OpenMRS Form containing this chart definition
-    public final List<ChartSection> tileGroups;
-    public final List<ChartSection> rowGroups;
+    public List<ChartSection> tileGroups;
```
capnfabs: Any reason for making these not `final`?

capnfabs: Oh, nevermind, I see it now.

capnfabs: Ok, now that I've actually read the code, I'd prefer if you kept these as `final`, and made `name` `final` and set as a constructor parameter. Keeping them `final` means that this class is immutable, which makes code significantly easier to reason about - you know that once a class is constructed, nothing is modifying values.

(Ideally, the `List`s would be immutable too, but one step at a time :smile:)

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -271,6 +281,28 @@ public void onPageFinished(WebView view, String url) {
...
+        List<Chart> charts = mController.getCharts();
+        String[] menuArray = new String[charts.size()];
+        int i = 0;
+        for (Chart chart : charts) {
+            menuArray[i] = chart.name;
+            i++;
+        }
+        mAdapter = new ArrayAdapter<String>(this, android.R.layout.simple_list_item_1, menuArray);
+        mDrawerList.setAdapter(mAdapter);
+        final Activity that = this;
```
capnfabs: I don't think this is used. But, for the future, if you need to access an outer class from within an inner class, you can use `OuterClass.this` from within the inner class.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -271,6 +281,28 @@ public void onPageFinished(WebView view, String url) {
...
             }
         });
+
+        initChartMenu();
+    }
+
+    private void initChartMenu() {
+        List<Chart> charts = mController.getCharts();
+        String[] menuArray = new String[charts.size()];
+        int i = 0;
```
capnfabs: If you have to maintain the loop variable anyway, you may as well use the 

``` java
for (int i = 0; i < charts.size(); i++) {
```

form.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -271,6 +281,28 @@ public void onPageFinished(WebView view, String url) {
...
+            menuArray[i] = chart.name;
+            i++;
+        }
+        mAdapter = new ArrayAdapter<String>(this, android.R.layout.simple_list_item_1, menuArray);
+        mDrawerList.setAdapter(mAdapter);
+        final Activity that = this;
+        mDrawerList.setOnItemClickListener(new AdapterView.OnItemClickListener() {
+            @Override
+            public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+                mController.updatePatientObsUi(position);
```
capnfabs: It's messy to pass the position through to the controller, and assume that the controller knows will look up the right chart to use. This is especially true because you're constructing the list items from the chart here, not in the controller, so the controller needs to know that this is a position in a list it otherwise has very little to do with.

Instead, could you pass a `Chart` object to the controller?

app/src/main/res/layout/fragment_patient_chart.xml
```diff
@@ -9,7 +9,7 @@
     OR CONDITIONS OF ANY KIND, either express or implied.  See the License for
     specific language governing permissions and limitations under the License.
 -->
-<RelativeLayout
+<android.support.v4.widget.DrawerLayout
```
capnfabs: I think you need to put a `RelativeLayout` as the first child in the `DrawerLayout`, like this:

``` xml
<DrawerLayout>
  <RelativeLayout>
    <!-- Content goes here -->
  </RelativeLayout>
  <ListView id="@+id/chart_list" / >
</DrawerLayout>
```

The `DrawerLayout` doesn't document that it works like a `RelativeLayout`, for all the content views you add, so we shouldn't assume that in the layout here. In fact, I'm surprised that the content renders correctly with this layout, because `DrawerLayout` inherits from `ViewGroup`, not from `RelativeLayout`.

app/src/main/res/layout/fragment_patient_chart.xml
```diff
@@ -86,4 +86,11 @@
       android:layout_height="fill_parent"
       android:layout_below="@+id/patient_chart_last_update_section"/>
 
-</RelativeLayout>
+  <ListView
+    android:id="@+id/chart_list"
+    android:layout_width="200dp"
```
capnfabs: can you specify this value and the background color in `dimens.xml` and `colors.xml` respectively?

app/src/main/res/layout/fragment_patient_chart.xml
```diff
@@ -86,4 +86,11 @@
...
+    android:layout_height="match_parent"
+    android:layout_gravity="left|start"
```
capnfabs: This should just be `start`. `left` and `start` mean the same thing for left-to-right languages, but they're a contradiction for right-to-left (RTL) languages. If we ever localise for RTL languages, we want the drawer to show up on the right-hand side :)

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
llvasconcellos: Yes, I know that. I did that just to test and forgot to rewrite the loop. I'll fix that.

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
llvasconcellos: Ooops. Forgot to remove this line. 
I didn't know about the OuterClass.this. I tried to use it inside the bellow onItemClick and it says: "Cannot resolve symbol". What I'm doing wrong?

llvasconcellos: OOOOhh, Now I get it: PatientChartActivity.this. 
Thanks for the tip!

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
llvasconcellos: The controller already handles the charts. I'm only building the list here because it's related to the view.
To change this is to start refactoring the this classes, something I was avoiding.

app/src/main/res/layout/fragment_patient_chart.xml
llvasconcellos: OK. Done it!

app/src/main/res/layout/fragment_patient_chart.xml
llvasconcellos: Done.

app/src/main/res/layout/fragment_patient_chart.xml
llvasconcellos: Done.

app/src/main/java/org/projectbuendia/client/models/Chart.java
llvasconcellos: Done!

capnfabs:
Ok to merge, thanks for the fixes. :+1:

capnfabs:
Ok to merge :+1:

PR #216, opened 2016-01-18
Send orderer UUID with each new order request from the client.

capnfabs:
This allows orders to be attributed to the user that created or edited them.


llvasconcellos:
Good to go!

PR #221, opened 2016-01-19
Clean up patient chart layout XML.

capnfabs:
- Remove an unused FrameLayout
- Fix padding on chart top row.


dancunningham:
Looks good to me!

PR #226, opened 2016-01-19
Fix Order Dosage and Frequency

### LINKED ISSUE #225: Treatment frequency no longer appearing
In the chart view it is meant to show both the Dosage and the Frequency on the second row in the chart.

This used to be working.

llvasconcellos:
This fixes #225


capnfabs:
Is this image from the linked #225 a screenshot of this fix?

Screenshot

If so, why is the medication name showing up twice?

llvasconcellos:
@capnfabs No, it isn't.

PR #228, opened 2016-01-20
Make start and stop (end) times for cell accessible via javascript

dancunningham:
In order to make the 5g/kg/day line work, the row script in the profile relies on being able to access the start and stop times for each grid cell. This makes those accessible via cell.dataset.start and cell.dataset.stop.


dancunningham:
Testing the fix now.

dancunningham:
Tested in wip build. This does work how I expected it to!

capnfabs:
I'm ok with this. Merge when ready.

PR #229, opened 2016-01-21
Add a notes panel to the patient chart screen.

capnfabs:
- Clean up adding encounters.
  - Previously, there was functionality to try and "guess" the type of observation. Remove this,
    the server has knowledge of concept type and can figure it out from the question UUID, and
    value type isn't used on the client.
  - Allow an `enterer_uuid` to be specified when an encounter is created, which allows the
    encounter to be attributed to a user.
  - Change representation of observations in an encounter from a Map of Question UUIDs --> Values
    to a List of JsonObservations. This means that we can process them consistently to the way
    the Observation syncing code does.
- Add a notes panel, which slides up from the bottom of the patient chart activity and allows notes
  to be entered by the user.


llvasconcellos:
Nothing to add. Good to merge.

PR #234, opened 2016-01-22
Trying to fix #199

### LINKED ISSUE #199: Crashes when form has empty sections
**Update:** looks like what is happening is that it doesn't like empty sections, which are caused by a section beginning with a select_multiple.

---

When I am opening the "Initial Physical Exam" form of profile "nutrition-itfc-101.csv" the app crashes as it is loading the form.

This is the profile that was loaded:
https://github.com/projectbuendia/profiles/blob/master/nutrition-itfc-101.csv

Here is a bug report:
https://www.dropbox.com/s/8ay3feu3grezfbs/bugreport-2016-01-14-00-29-13.txt?dl=0

App build 900 (revision #3d9fc283d508b936794e50a3dea0f6004ec96e32).

Connected to dev.projectbuendia.org

Patient ABC/680 "Chinasa Tremblay" in location "UNTI Soins Intensifs".

llvasconcellos:
Solution to infinite recursion.
**\* DO NOT MERGE ***
For @capnfabs!

### DIFF
--- a/third_party/odkcollect/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
+++ b/third_party/odkcollect/src/main/java/org/odk/collect/android/activities/FormEntryActivity.java
@@ -1448,13 +1448,13 @@ public class FormEntryActivity
                         && preset.targetGroup.equals(groups[groups.length - 1].getLongText())) {
                     mTargetView = odkv;
                 }
-				Log.i(TAG,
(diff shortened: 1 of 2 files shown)


capnfabs:
Thanks for helping with this @llvasconcellos! Got a fix in #237, so I'm going to close this PR.

PR #236, opened 2016-01-22
Tweak notes panel to take up less space

capnfabs:
Before:

https://drive.google.com/file/d/0B0LJ7nmd6PoyM194emY1MU85NjA/view

After:

https://drive.google.com/file/d/0B0LJ7nmd6PoyV21nbUZwejZtUnM/view

Some things that I'd like your thoughts on:

1 - Background shading. I don't think we need it, the grey background on the notes panel provides enough contrast and the shading makes the chart difficult to read.
2 - The size of the notes panel. It's at 50% now as opposed to 70% earlier.
3 - The notes panel now stays the same size when the keyboard expands. This looks pretty bad at 50% because there's no room for old notes, but could work well at larger sizes.

My preference is:
1. Eliminate background shading, it's not necessary.
2. Stick with 70% notes panel height. At 50%, it's impossible to read old notes whilst writing new notes. I could be talked down to 65% maybe :)
3. I like that the notes panel stays the same size when the keyboard pops up, I think we should keep this.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
@@ -99,7 +99,7 @@ public final class PatientChartActivity extends BaseLoggedInActivity {
 
     private static final String KEY_CONTROLLER_STATE = "controllerState";
     private static final String SEPARATOR_DOT = "\u00a0\u00a0\u00b7\u00a0\u00a0";
-    private static final float PANEL_HEIGHT_FRAC = 0.5f;
+    private static final float PANEL_HEIGHT_FRAC = 0.6f;
 
     private PatientChartController mController;
     private boolean mIsFetchingXform = false;
@@ -125,6 +125,7 @@ public final class PatientChartActivity extends BaseLoggedInActivity {
     @InjectView(R.id.notes_panel_text_entry) EditText mAddNoteEntryText;
     @InjectView(R.id.notes_panel_btn_save) View mAddNoteButton;
     @InjectView(R.id.notes_panel_submit_spinner) View mAddNoteWaitingSpinner;
+    @InjectView(R.id.patient_chart_container) View mPatientChartContainer;
 
     private static final String EN_DASH = "\u2013";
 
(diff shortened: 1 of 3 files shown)


capnfabs:
WIP build here: http://jenkins.projectbuendia.org/job/%7Eclient-wip/57/

dancunningham:
Have watched the videos + tried it out on wip.
1. Agreed
2. I think 50% is a lot more usable.
3. Definitely squashes the notes into too tiny a space keeping it at 50% when the keyboard is up. I think it feels better for its size to increase by the height of the keyboard, so you see the same notes as you did before. But maybe we don't have space for that /and/ being able to see much of the chart.

Try 60% fixed height? (final offer!)

Two more additional tweaks:

A. Change "No recorded observations" to "No recorded notes"
B. It should start off scrolled to the bottom of the notes, so you see the most recent one (similarly after you add a note, it should scroll to the bottom so you see that your note was added)

dancunningham:
Found a bug!!!

You cannot scroll to see the bottom of the chart when the notes panel is open.

Because it's scrolling underneath, rather than squashing the size of the webview.

Proper fix is probably to dynamically resize the webview container to fit the space left when notes panel is up.

Hackier fix might be just to pad the bottom of the webview with sufficient padding to allow you to scroll to the bottom.

dancunningham:
One more: the "Add Treatment" button is hidden under the Notes panel even when it is in its collapsed state.

capnfabs:
By default, the SlidingUpPanelThingy takes care of all this, but I turned on a setting that I shouldn't have in this PR. The fix is to remove the line:

``` xml
app:umanoOverlay="true"
```

capnfabs:
Ok, fixes and tweaks are uploaded. @llvasconcellos, mind taking a look?

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
```diff
@@ -351,6 +381,44 @@ public void onClick(View v) {
...
+     *    new parent height).
+     *  - For small values of layout_weight (e.g. < 0.5) this makes the notes panel noticeably
+     *    larger when the keyboard is expanded.
+     *  - So we compute the height at a time when we know the keyboard isn't showing, and assign
+     *    (weight * parent height) to the margin, so that much of the underlying view will always
+     *    be visible.
+     *
+     * @param panelHeightFrac the height to which the notes panel should be set at.
+     */
+    private void setNotesPanelMaxHeightToFracOfParent(final float panelHeightFrac) {
```
llvasconcellos, line 398: Great. Thinking in different screen sizes. :+1:

llvasconcellos, line 398: Android layout quirks. There isn't a way of doing this using %? (Just out of curiosity)

llvasconcellos:
@capnfabs Good to go!

app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
capnfabs, line 398: Yeah, there's a bunch of issues here.
1 - We have to resize the WebView dynamically because the `SlidingUpPanelLayout` doesn't measure the child views when the panel expands and contracts. This is an important performance optimization - measure / layout is expensive.
2 - The IME causes the Window to resize, so if we use a percentage, it's a percentage of the available space when the IME is both open and closed. The actual number of pixels is different though, so you end up with this weird thing where the notes panel resizes by a small amount when the keyboard opens.

PR #237, opened 2016-01-22
Allow empty sections in ODK forms.

### LINKED ISSUE #199: Crashes when form has empty sections

capnfabs:
Fixes #199.
- In FormEntryActivity, remove the try{} block that catches RuntimeExceptions and tries to skip
  forward. It just hides errors.
- In FormController, remove assumption that groups are non-empty, and update the "next entry" logic
  to handle empty groups.

### DIFF
--- a/third_party/odkcollect/src/main/java/org/odk/collect/android/logic/FormController.java
+++ b/third_party/odkcollect/src/main/java/org/odk/collect/android/logic/FormController.java
@@ -599,8 +599,13 @@ public class FormController {
         }
 
         // jump to the end of the group
-        mFormEntryController.jumpToIndex(indicies.get(indicies.size() - 1));
-        return stepToNextEvent(STEP_OVER_GROUP);
+        if (indicies.size() > 0) {
+            mFormEntryController.jumpToIndex(indicies.get(indicies.size() - 1));
+            return stepToNextEvent(STEP_OVER_GROUP);
+        } else {
+            FormIndex idx = mFormEntryController.getModel().incrementIndex(getFormIndex());
+            return mFormEntryController.jumpToIndex(idx);
+        }
     }
 
     /**
@@ -912,7 +917,7 @@ public class FormController {
 
         // For questions, there is only one.
         // For groups, there could be many, but we set that below
-        FormEntryPrompt[] questions = new FormEntryPrompt[1];
+        FormEntryPrompt[] questions = new FormEntryPrompt[0];
 
     	IFormElement element = mFormEntryController.getModel().getForm().getChild(currentIndex);
         if (element instanceof GroupDef) {
@@ -965,7 +970,7 @@ public class FormController {
             }
         } else {
             // We have a quesion, so just get the one prompt
-            questions[0] = getQuestionPrompt();
+            questions = new FormEntryPrompt[] {getQuestionPrompt()};
         }
 
         return questions;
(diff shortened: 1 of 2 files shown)


capnfabs:
Screenshot from the `nutrition-itfc-101.csv` profile that wasn't working from a week or two ago:

dancunningham:
+10 for the screenshot annotation!

third_party/odkcollect/src/main/java/org/odk/collect/android/logic/FormController.java
```diff
@@ -599,8 +599,13 @@ private int stepOverGroup() {
         }
 
         // jump to the end of the group
-        mFormEntryController.jumpToIndex(indicies.get(indicies.size() - 1));
-        return stepToNextEvent(STEP_OVER_GROUP);
+        if (indicies.size() > 0) {
```
dancunningham, line 602: correct spelling is "indices" I think?

capnfabs, line 602: Indeed it is! But this isn't our code so I'm going to leave it as is.

dancunningham, line 602: OK! Yeah probably best not break things just for the sake of spelling :-)

dancunningham:
Been looking through the code (so much code!) to try to figure out if there's an easy way to style the multi_select question headers just like normal questions rather than like section headers to fix the visual hierarchy here. Can't find where I might change that though, so I'm guessing it's not an easy enhancement to make?

third_party/odkcollect/src/main/java/org/odk/collect/android/logic/FormController.java
```diff
@@ -599,8 +599,13 @@ private int stepOverGroup() {
...
+            mFormEntryController.jumpToIndex(indicies.get(indicies.size() - 1));
+            return stepToNextEvent(STEP_OVER_GROUP);
+        } else {
+            FormIndex idx = mFormEntryController.getModel().incrementIndex(getFormIndex());
+            return mFormEntryController.jumpToIndex(idx);
```
llvasconcellos, line 607: Well done!!
That's what I was trying to do.

llvasconcellos:
@capnfabs Just one thing that I noticed is that the Section header is the same size of a normal question and it should be bigger.

llvasconcellos:
If the size is not a problem this solution looks good to merge!

capnfabs:
So, fun fact - the section headers have always been the same size as the select_multiple questions, but the other questions have smaller headers.

The reason is because select_multiple actually creates a top-level section, and then the radio buttons underneath are the questions.

Initially, I tried to fix this problem by parenting the select_multiple questions under the actual sections, but ODKCollect can't handle nested sections, and so it didn't work. So, we're stuck with the weird heading size problem for the time being.

capnfabs:
So @dancunningham , to answer your question - no. It's really hard :)

llvasconcellos:
@capnfabs I noticed that too.

PR #238, opened 2016-01-22
Fix test compilation

capnfabs:
These META-INF files got included in both `espresso` and `espresso-web` `classes.jar` files somehow, which breaks test suite compilation. See [Espresso-web import causes duplicateFileException](http://stackoverflow.com/questions/33800924/espresso-web-import-causes-duplicatefileexception) for more details.


llvasconcellos:
Great! Good to go!

PR #239, opened 2016-01-22
Only refresh the patient chart if something changed after a sync.

capnfabs:
Previously, we refreshed the patient chart whenever a sync completed, even if there was no
additional data. Now, we only refresh the chart if the data has changed.

Note that we could still improve upon this a lot - for example, by checking for whether observations
for the given patient had changed - but our data model doesn't support this at the moment.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
@@ -8,12 +8,9 @@ import android.webkit.WebView;
 import com.google.common.collect.Lists;
 import com.mitchellbosecke.pebble.PebbleEngine;
 
-import org.joda.time.Chronology;
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.joda.time.ReadableInstant;
-import org.joda.time.chrono.ISOChronology;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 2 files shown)


dancunningham:
YES!!!!

dancunningham:
The refreshing causes a slight hang in the UI, this is a brilliant amendment. Thank you!

llvasconcellos:
\o/ wooohooo! God to merge!
Just have a look at this:
https://github.com/projectbuendia/client/blob/dev/app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java#L90

capnfabs:
Thanks @llvasconcellos, I hadn't noticed that. I wonder why it wasn't working? I think the code you're pointing to is a better solution, so I'll try and figure out what's going on there and fix that instead of merging this change.

llvasconcellos:
@capnfabs :+1:

capnfabs:
I fixed the root cause in the latest commit. Please take a look.

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -36,6 +33,7 @@
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
llvasconcellos, line 36: @capnfabs Do you really need this import?

llvasconcellos:
Great! Good to go!

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
capnfabs, line 36: No, because we could use `lastChart.equals(chart.name)` below as before, but I like `Objects.equals` for non-final variables because it's `null`-safe as well :)

PR #241, opened 2016-01-24
Fix Required Fields

### LINKED ISSUE #89: required=(blank) not working
The intended behaviour when required is left blank is that if a field has not been filled in ever for that patient, that row does not appear on the patient chart. It should only appear if it has ever been entered.

If required=yes, it will always appear even if it's never been entered.

For select_one fields, when required is left blank, the row is appearing even when nothing has ever been entered.

This is with build b7b1b5b.

UPDATE: This now seems to be not working for any type of field (select_one, select_multiple, number)

llvasconcellos:
This fixes #89
Needs to merge server side as well:
https://github.com/projectbuendia/buendia/pull/70

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -7,18 +7,6 @@
 <script src="jquery-1.5.1.min.js"></script>
 <script src="freezeheader.js"></script>
 <script src="chart.js"></script>
-<script>
-  function isRowEmpty(parentNode) {
-    var rowEmpty = true;
-    $(parentNode).find('td').each(function(index, element) {
-      var innerHtml = element.innerHTML.trim();
-      if ( innerHtml != "" ) {
-        rowEmpty = false;
-      }
-    });
-    return rowEmpty;
-  }
-</script>
 
 <!-- Use double-quotes for HTML attributes; single-quotes for JS/Pebble strings. -->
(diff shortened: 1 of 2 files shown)


app/src/main/assets/chart.html
```diff
@@ -7,6 +7,18 @@
 <script src="jquery-1.5.1.min.js"></script>
 <script src="freezeheader.js"></script>
 <script src="chart.js"></script>
+<script>
+  function isRowEmpty(parentNode) {
```
capnfabs: This should probably go in `chart.js`

capnfabs: Actually, can we do this in pebble instead of in JS? It would fit better with our current programming model that way, and I think it would be more performant.

llvasconcellos: @capnfabs I tried but pebble worn me out. There is a bug regarding variable scope in pebble and that's why I choose this solution. We really need to revome pebble as soon as possible.

llvasconcellos: @capnfabs Trust me, that wasn't my first option.

llvasconcellos: Done!

capnfabs:
:+1: ready to merge

PR #243, opened 2016-01-25
Enable like search.

### LINKED ISSUE #176: Fix "Search by ID" to cope with ID prefixes
It should match on just the ID (numerical) part

llvasconcellos:
This fixes #176 

Must merge this PR as well:
https://github.com/projectbuendia/buendia/pull/71
https://github.com/projectbuendia/db-snapshot/pull/2


capnfabs:
I'm not sure where to comment on this set of changes; sorry if this is the wrong place!

My concern is that strings like '123' will also match strings like '451236'. Do you think it's worth addressing this?

llvasconcellos:
@capnfabs yes, they will match but i think it still useful. We can probably merge as it is and think of a better way for improving it if it feels like need of improvement.
At least its working now.

capnfabs:
Ok! This client-side change looks ok to me then. :+1:

PR #248, opened 2016-01-27
UI support for setting custom start dates on Orders.

capnfabs:
- You can set the start date when you first create an order, but not when you edit it.
- Fix the "stop after X days" logic to be a straightforward duration since start date, instead of being the duration from today. This makes a lot more sense now that we're actually displaying the start date.


llvasconcellos:
Reviewed it. Built it. Love it.

PR #255, opened 2016-01-28
Fix local storage of encounter times.

capnfabs:
ODK rounds times to the nearest minute for submission. We work around this by sending the real
submission time (including seconds) in JSON with the completed Xform. The temporary observations
stored on the client until the next sync, however, are given the timestamp extracted from the Xform,
which means that under the following circumstances:
- An Xform is filled out at 12:03:05
- The values are added as temporary observations, with the timestamp 12:03:00
- A sync completes at 12:03:08, which causes those temporary observations to be replaced with the
  real observations, timestamped correctly at 12:03:05
- Another Xform is filled out at 12:03:25
- The values are added as temporary observations, with the timestamp 12:03:00
- These values do not show up on the patient chart, because the previously synced older values have
  a later timestamp.

This change fixes that problem by using the actual time of submission to store temporary
observations, instead of the time extracted from the Xform.

This is a partial fix for #125.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
@@ -31,6 +31,7 @@ import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.xform.parse.XFormParser;
 import org.joda.time.DateTime;
 import org.joda.time.format.ISODateTimeFormat;
+import org.json.JSONException;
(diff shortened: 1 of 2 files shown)


capnfabs:
Note: there are also a bunch of formatting changes in this PR. I left them in because the auto-formatter in Android Studio keeps trying to fix style issues in this code, and I got sick of resisting it.

capnfabs:
I've got a better approach incoming.

PR #256, opened 2016-01-28
Update observations from Xform submission response.

### LINKED ISSUE #125: PatientChartActivity drops temporary observations prematurely.

capnfabs:
Previously, when submitting an Xform, we parsed the XML in the request body to add a set of
"temporary observations" to the database. These had the correct values, but didn't have UUIDs from
the server.

This mechanism was prone to the issues described in #125, so instead of using temporary
observations, we retrieve the result of the Xform submission from the server, and insert real,
server-provided observation information into the local data store instead.

Fixes #125.

_Note_: this change also includes some formatting fixes, I got sick of fighting the Android Studio auto-formatter.

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
@@ -24,27 +24,25 @@ import com.android.volley.Response;
 import com.android.volley.TimeoutError;
 import com.android.volley.VolleyError;
 import com.google.common.base.Charsets;
-import com.google.common.base.Joiner;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 
-import org.javarosa.core.model.data.IAnswerData;
-import org.javarosa.core.model.instance.TreeElement;
-import org.javarosa.xform.parse.XFormParser;
-import org.joda.time.DateTime;
-import org.joda.time.format.ISODateTimeFormat;
 import org.json.JSONObject;
 import org.odk.collect.android.activities.FormEntryActivity;
 import org.odk.collect.android.application.Collect;
 import org.odk.collect.android.model.Preset;
 import org.odk.collect.android.provider.FormsProviderAPI;
 import org.odk.collect.android.tasks.DeleteInstancesTask;
-import org.odk.collect.android.utilities.FileUtils;
 import org.projectbuendia.client.App;
 import org.projectbuendia.client.AppSettings;
 import org.projectbuendia.client.events.FetchXformFailedEvent;
 import org.projectbuendia.client.events.SubmitXformFailedEvent;
 import org.projectbuendia.client.events.SubmitXformSucceededEvent;
 import org.projectbuendia.client.exception.ValidationException;
+import org.projectbuendia.client.json.JsonEncounter;
 import org.projectbuendia.client.json.JsonUser;
+import org.projectbuendia.client.json.Serializers;
+import org.projectbuendia.client.models.Encounter;
 import org.projectbuendia.client.net.OdkDatabase;
 import org.projectbuendia.client.net.OdkXformSyncTask;
 import org.projectbuendia.client.net.OpenMrsXformIndexEntry;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -87,17 +79,19 @@
...
-                    for (OpenMrsXformIndexEntry formEntry : response) {
-                        fetchAndCacheXForm(formEntry);
+                new Response.Listener<List<OpenMrsXformIndexEntry>>() {
+                    @Override
+                    public void onResponse(final List<OpenMrsXformIndexEntry> response) {
+                        for (OpenMrsXformIndexEntry formEntry : response) {
+                            fetchAndCacheXForm(formEntry);
+                        }
+                    }
+                }, new Response.ErrorListener() {
```
llvasconcellos, line 100: @capnfabs It's difficult to use the Generic error listener (the one with the toasts)?

llvasconcellos, line 100: Well, don't need to do it right now though.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -546,7 +538,7 @@ private static String readFromPath(String path) {
             }
             return sb.toString();
         } catch (IOException e) {
-            LOG.e(e, format("Failed to read xml form into a String. FilePath=  ", path));
+            LOG.e(e, format("Failed to read xml form into a String. FilePath=  %s", path));
```
llvasconcellos, line 552: Good catch.

llvasconcellos:
@capnfabs Looks good.

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
```diff
@@ -554,166 +546,19 @@ private static String readFromPath(String path) {
...
+        // TODO: don't parse this here or do a roundtrip text --> JSON --> text --> GSON conversion
+        GsonBuilder gsonBuilder = new GsonBuilder();
+        Serializers.registerTo(gsonBuilder);
+        Gson gson = gsonBuilder.create();
+        JsonEncounter jsonEncounter = gson.fromJson(response.toString(), JsonEncounter.class);
+        Encounter encounter = Encounter.fromJson(jsonEncounter.patient_uuid, jsonEncounter);
+        ContentValues[] values = encounter.toContentValuesArray();
+        if (values.length > 0) {
+            resolver.bulkInsert(Contracts.Observations.CONTENT_URI, values);
+
```
viniciusboson, line 569: Awesome refactoring, @capnfabs. It helps fixing #125 in a such cleaner and clever way. Congrats.

viniciusboson:
Fantastic work @capnfabs.
These changes may fix the same error found in #172, right ?

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
capnfabs, line 100: I don't think it would be difficult, it was just like this when I got here :)

app/src/main/java/org/projectbuendia/client/ui/OdkActivityLauncher.java
capnfabs, line 552: I had a lint warning :)

capnfabs:
Hey @viniciusboson! Yeah, I think they'll fix the same problems that you're dealing with, so long as all the forms have definitely finished submitting before sync starts. There might still be a "flicker" where an observation exists, and then disappears, and then comes back again if a sync occurs before the form is uploaded, however. So it's important to make sure that all forms are submitted before sync proceeds :)

I realised overnight that your code still depends on a lot of the temporary observations logic that I deleted in this change, so I'll add that logic back in now, so it's not deleted when I merge this PR.

viniciusboson:
Hey @capnfabs,
As far as i understood, these changes that you reverted could work fine with my PR too.
I would just need to merge and check if it still works ...
I really liked what you did ..

PR #257, opened 2016-01-28
Use dark, ancient magic to ensure that seconds are set for form entries.

### LINKED ISSUE #125: PatientChartActivity drops temporary observations prematurely.

capnfabs:
The timestamp for encounters entered with an Xform didn't include a seconds field previously. This meant that if the same form was filled out twice in one minute, there was no guarantee that the most recently set value would be the one displayed on the patient chart.

This change makes the DateTimeWidget in Xforms use the seconds of the current time by default, and then clears the seconds if the value changes. This workaround gets us a lot closer to the behaviour that the user expects.

Fixes #125.

### DIFF
--- a/third_party/odkcollect/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java
+++ b/third_party/odkcollect/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java
@@ -52,6 +52,7 @@ public class DateTimeWidget extends QuestionWidget {
     private boolean hideMonth = false;
     private boolean showCalendar = false;
 	private HorizontalScrollView scrollView = null;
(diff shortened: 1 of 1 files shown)


llvasconcellos:
I love the title! OK. Good to go!

PR #258, opened 2016-01-29
Persistent fields

dancunningham:
This is designed so that on opening the Diagnosis form, the previously selected Diagnoses and Complications are selected by default so you just need to change the ones you want to select and unselect.

Less hacky than I was expecting actually and I refactored some bits as I did it.

All the nasty hardcoded concept IDs and labels are confined to one place.

Various TODOs added as appropriate.


app/src/main/java/org/projectbuendia/client/models/ConceptUuids.java
```diff
@@ -25,6 +26,132 @@
...
+        PERSISTENT_FIELDS.put("777130473AAAAAAAAAAAAAAAAAAAAAAAAAAA", "Recurrent oedema");
+        PERSISTENT_FIELDS.put("1777000638AAAAAAAAAAAAAAAAAAAAAAAAAA", "Persistent fever");
+        PERSISTENT_FIELDS.put("1777000639AAAAAAAAAAAAAAAAAAAAAAAAAA", "Refeeding syndrome");
+        PERSISTENT_FIELDS.put("1777000640AAAAAAAAAAAAAAAAAAAAAAAAAA", "Difficulties with breastfeeding");
+        PERSISTENT_FIELDS.put("777135595AAAAAAAAAAAAAAAAAAAAAAAAAAA", "Persistent lack of appetite");
+        PERSISTENT_FIELDS.put("1777000641AAAAAAAAAAAAAAAAAAAAAAAAAA", "inappropriate weight response (if stagnant or losing weight in phase 2)");
+        PERSISTENT_FIELDS.put("1777000642AAAAAAAAAAAAAAAAAAAAAAAAAA", "child behavioural problems");
+    }
+
+    public static final String PHASE_UUID = "UUID";
```
capnfabs: What are these? It doesn't look like you're using them.

app/src/main/java/org/projectbuendia/client/models/ConceptUuids.java
```diff
@@ -25,6 +26,132 @@
     public static final String HYDRATION_UUID = "162653AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
```
capnfabs: Are any of these unused now? If so, please clean them up. You can check in Android Studio.

capnfabs:
Looks good aside from the comments I added. I'd advise not merging this until we can support fetching the labels from the synced concepts, considering this change isn't useful for the pilot until it works in French.

capnfabs:
Also congrats on your first substantial Android change :)

capnfabs:
Ok to submit :+1: nice work.

llvasconcellos:
wow! Very nice!

PR #259, opened 2016-02-01
Fixes #177: "Sorry, form save failed: Index 0 requested..."

### LINKED ISSUE #177: Admission and History form: Sorry, form save failed: Index 0 requested, with a size of 0
Trying to save admissions form, getting toast “Sorry, form save failed: Index 0 requested, with a size of 0”
(on build 880 connected to dev)

Anthropometrics form and Intensive Observations form work.

capnfabs:
ODK keys its form type data off of an `_ID` column, which is a primary key on the `forms` table.
This `_ID`, however, isn't explicitly set anywhere - it's an auto-incrementing primary key.

When a sync occurred whilst a form was being filled out, the form was being replaced in the ODK
database. ODK's FormsProvider, however, models REPLACE operations as DELETEs followed by INSERTs in
a single transaction. This meant that that the `_ID` column changed for every form on every sync,
which means that when ODK goes to look up the form information again on submit, the relevant form
had disappeared.

This change fixes the problem by figuring out the `_ID` of the existing record, and then doing an
`INSERT OR REPLACE` instead of the transactional `DELETE` and `INSERT`.

A better long-term fix might be to make the ODK keep the form information in memory whilst the form
was being filled out, but that's fiddlier.

### DIFF
--- a/third_party/odkcollect/src/main/java/org/odk/collect/android/provider/FormsProvider.java
+++ b/third_party/odkcollect/src/main/java/org/odk/collect/android/provider/FormsProvider.java
@@ -379,16 +379,35 @@ public class FormsProvider extends ContentProvider {
 
 		SQLiteDatabase db = getDbHelper().getWritableDatabase();
 
-        long rowId;
-        if (replace) {
-            // If we want this insert to always replace, then do a transactional delete followed
-            // by an insert. The primary key is _ID, which we can't know in advance.
-            try {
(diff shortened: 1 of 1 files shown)


capnfabs:
@llvasconcellos Please sanity-check this when you have time, but I'll merge it now.

llvasconcellos:
Clean fix! Nothing to add.

PR #260, opened 2016-02-06
Add a script that makes a "release-ish" build.

capnfabs:
This build has the same parameters as a release build - require WiFi, production username / password
defaults set - but is a debug build, because our "release build" process is difficult when there's
nobody really sure of how it works anymore. Also, it's useful to be able to get a tablet with a bug
and debug it live in the field.

Note that you need to add a file named `.openmrs_password` that only contains the OpenMRS password.
This file is explicitly ignored by git to avoid storing the password in source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ local.properties
 .classpath
 .settings
 .project
+
+# OpenMRS password file.
+.openmrs_password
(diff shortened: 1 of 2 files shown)


llvasconcellos:
:+1:

PR #261, opened 2016-02-06
Add a "Stop order now" button.

capnfabs:
Note that I had to make some changes to support when an order gets a duration of zero, in a
circumstance when (e.g.) a user creates an order, administers the order, and then stops the order.
These changes are hacky at the moment because our data model doesn't support < 1 day duration
orders. Long term, we should clean this up, but we have an outstanding bug for that anyway.

Also: Cleanup order fragment XML.
- Remove unnecessary LinearLayout
- Add `android:labelFor` attributes for TextViews that point at EditTexts.

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -135,6 +135,15 @@
           {% for column in columns %}
             <td class="{{column.start == nowColumnStart ? 'now' : ''}}">
               {% set active = intervals_overlap(order.interval, column.interval) %}
+              <!--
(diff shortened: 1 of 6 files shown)


capnfabs:
This is a pretty hacky change. I'm not sure if it's worth merging in this form / in its entirety, so I've left it open for the time being. Note that this has been cherry-picked onto the pilot-feb-2016 branch though.

llvasconcellos:
I'll look at it carefully.

capnfabs:
Too hacky, not merging.

PR #262, opened 2016-02-06
Fix the bug where observations don't reliably when submitting forms.

capnfabs:
- The server-side logic that was supposed to do this doesn't work on the edisons due to a timing
  issue. This change re-introduces the temporary cache in the event that the client doesn't obtain
  anything from the server, which should make the problem go away (symptomatically at least).

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/sync/controllers/ObservationsSyncPhaseRunnable.java
+++ b/app/src/main/java/org/projectbuendia/client/sync/controllers/ObservationsSyncPhaseRunnable.java
@@ -13,13 +13,10 @@
 
 package org.projectbuendia.client.sync.controllers;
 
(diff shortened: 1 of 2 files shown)


capnfabs:
Merging because this will be useful in dev.

llvasconcellos:
I don't quite understand what's going on here but code wise everything seems fine.

capnfabs:
Yeah, this one was super complicated. This is the way that we used to do it, but then I thought we could move the hack over to the server side to cleanup the client side code. That worked in my local testing, but didn't work on the Edisons, because Xforms take a few seconds to process on the Edisons, and they process asynchronously - so there's no way for us to consistently retrieve the created encounter on the server. That's why I restored this client-side workaround instead.

PR #265, opened 2016-02-08
Disable the observations popup for tiles with multiple concepts.

capnfabs:
They misrepresent the information by only displaying one concept, so we just disable them entirely
for the time being.


llvasconcellos:
:+1:

PR #267, opened 2016-02-08
Load title for Notes Panel from concept dictionary.

capnfabs:
This allows the Notes panel to be used for other purposes.


llvasconcellos:
:+1:

PR #268, opened 2016-02-08
Fix "Void Observations" so that it actually works.

capnfabs:
Previously, it deleted the observation locally, and made a request to an endpoint that didn't exist,
but didn't do anything with the error anyway, so there was no real indication that it wasn't
working.

This change fixes all that, and also:
- Introduces a new OpenMrsStringRequest, which allows for empty responses like 204s (as opposed to
  OpenMrsJsonRequest)
- Pops a toast when an observation isn't successfully deleted
- Also pops a toast when a treatment (order) isn't successfully deleted, which we weren't doing
  either
- Adds padding around the "Void Observation" toggle switch so that it's easier to press without
  accidentally closing the dialog.


llvasconcellos:
Wow, lot's of changes! Great work!

PR #270, opened 2016-02-08
Fix a crash relating to displaying forms with persistent fields.

capnfabs:
If a patient has observations that use question UUIDs that aren't in our forms or charts, we don't download the concept. When we try and load the concept name to tell ODK collect to persist the field, it doesn't work because the concept UUID is null.

We fix that with a null check.


llvasconcellos:
:+1:

PR #271, opened 2016-02-09
Restore clickability for all tiles with 1 concept.

capnfabs:
Previously, all tiles had their clicking disabled by faulty logic. Pebble doesn't know how to handle
the `.length` property of arrays, and interpreted it as null, which prevented the relevant code
from firing. This change converts `ChartItem#conceptUuids` to a List instead of an array, such that
Pebble is capable of working with the field.


llvasconcellos:
Let's ditch peeble please!!! :smile:

llvasconcellos:
great work!

PR #272, opened 2016-02-10
Show weight at top

dancunningham:
Remove "Days since Symptom onset" and replace it with "Weight" in the top section.

Weight is a critical metric for clinical management of a malnourished patient and determines things like how much to feed and dosages of medication so it's very useful that it's in the top portion of the patient chart which is frozen to the top when you scroll down to the feeding and treatment sections of the chart.


app/src/main/res/values/strings.xml
```diff
@@ -142,6 +142,7 @@
 
   <string name="since_admission">Since admission</string>
   <string name="since_symptom_onset">Since symptom onset</string>
+  <string name="weight">Weight</string>
```
capnfabs, line 145: Should probably be Weight (kg).

Long term, I think we'd want to load these strings from the concept name in the profile.

capnfabs:
Ok to merge :+1:

llvasconcellos:
:+1:

PR #275, opened 2016-02-17
[2016] Don't repeatedly attempt login with an incorrect password

### LINKED ISSUE #106: Incorrect password causes lockout

llvasconcellos:
This fixes #106


capnfabs:
Nice work.

capnfabs:
This is ready to merge.

PR #276, opened 2016-02-28
[2016] Configurable chart column interval

llvasconcellos:
This is the client side of this pivotal story: https://www.pivotaltracker.com/story/show/111137168
The server part is here: https://github.com/projectbuendia/buendia/pull/89

This changes allow to specify the column interval for the chart. The configuration should be added to the format column in the profile csv.
Examples: 

> {column_type: 'daily'}
> {column_type: 'timed', column_time: 30}
> {column_type: 'encounter', column_time: 1}

The column_type accept the following 3 values:
- **daily**: one column for each day (previous behavior)
- **timed**: columns will be rendered following the column_time interval. One day will have many columns depending on the column_time in minutes.
- **encounter**: column will be inserted in the time of each encounter but merged according to column_time interval in minutes.


app/src/main/java/org/projectbuendia/client/models/Chart.java
```diff
@@ -11,20 +11,40 @@
...
+
     public final String uuid;  // UUID of the OpenMRS Form containing this chart definition
     public final List<ChartSection> tileGroups;
     public final List<ChartSection> rowGroups;
     public final String name;
+    public String columnType;
+    public int columnTime;
 
-    public Chart(String uuid, String name) {
+    public Chart(String uuid, String name, String format) {
```
capnfabs, line 32: Please document the format of the `format` parameter.

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -191,18 +210,71 @@ void addOrders(List<Order> orders) {
...
+                default:
+                    returnColumn =  getDailyColumn(instant);
+                    break;
+            }
+            return returnColumn;
+        }
+
+        Column getDailyColumn(ReadableInstant instant) {
+            LocalDateTime date = new DateTime(instant).toLocalDateTime();  // a day in the local time zone
+            DateTime start = date.withTime(0, 0, 0, 0).toDateTime();
```
capnfabs, line 231: I think you can use `LocalDateTime.withTimeAtStartOfDay` instead of `withTime(0,0,0,0)`

app/src/main/java/org/projectbuendia/client/ui/chart/ChartRenderer.java
```diff
@@ -191,18 +210,71 @@ void addOrders(List<Order> orders) {
...
                 String dateLabel = date.toString("d MMM");
-                mColumnsByStartMillis.put(startMillis, new Column(
-                    start, start.plusDays(1), admitDayLabel + "<br>" + dateLabel));
+                mColumnsByStartMillis.put(columnInterval, new Column(
+                    columnInterval, admitDayLabel + "<br/>" + dateLabel));
+            }
+            return mColumnsByStartMillis.get(columnInterval);
+        }
+
+        Column getEncounterColumn(ReadableInstant instant) {
```
capnfabs, line 248: For `column_type: 'encounter'`, I can't work out what the `column_time` parameter actually does. To me, it looks like it doesn't make a difference to the way that the columns are computed, and each encounter will have a new column created for it. Could you clarify your design for me a bit please?

capnfabs:
Nice work! You have a good understanding of the code here.

PR #294, opened 2019-06-17
Show a selectable number of columns per day.

zestyping:
Issues: none
Scope: chart activity, chart renderer, chart HTML and CSS

#### User-visible changes

A new "Zoom" button pops up a dialog that allows you to select 1, 2, 3, or 4 columns per day.  When more than 1 column is shown per day, the time span of each column is shown in a second row of column headers.

To improve visual understanding, gaps that represent omitted (empty) columns are drawn with grey diagonal stripes.

### DIFF
--- a/app/src/main/assets/chart.css
+++ b/app/src/main/assets/chart.css
@@ -10,16 +10,32 @@ th, td {
   text-align: left;
   vertical-align: baseline;
 }
+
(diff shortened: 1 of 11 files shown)


zestyping:
Some examples of the new layout:

2colperday

3colperday

PR #318, opened 2019-06-20
Switch over to using demo.buendia.org as the default server.

zestyping:
Issues: none
Scope: build.gradle
Requested reviewers: @ivangayton


ivangayton (approved):
Fine, merging.

PR #331, opened 2019-06-24
Krtonga/remove update chart from dropdown

krtonga:
I am removing the update chart option from the dropdown menu.

Note: the blank line in the dropdown menu was appearing because the `BaseLoggedInActivity` adds a user profile icon to the menubar. The user item has no title, and the click listener was not working because the icon was not designed to be in a dropdown. As a simple fix, I set `showAsAction` to `always`. Please let me know if a better solution would be to not show this menu item on this page.


zestyping (approved):
This looks great!  Thanks!

Issue #325, opened 2019-06-21
Patient Record -> If long name, Age and Sex not displayed


saratheneale:
In the Patient Record, if the name is long, the Age and Sex are not displayed in the Header Bar. This seems useful information to always show with higher priority than the name. 

Picture with short name. Notice the Sex and Age are visible. 
![20190621_151114](https://user-images.githubusercontent.com/3619338/59953966-5e6fad00-9437-11e9-86a7-b32fab849f1f.jpg)

Picture with Long name. Notice Sex and Age are NOT visible: 
![20190621_151058](https://user-images.githubusercontent.com/3619338/59953985-6deef600-9437-11e9-8efb-85df9cc118cd.jpg)

PR #332, opened 2019-06-25
Display patient sex and age in subtitle on PatientChartActivity

mjanes:
Solution to https://github.com/projectbuendia/client/issues/325

This does leave display of patient and sex and age in smaller font than previously, even if id and name is short. I am uncertain if that is acceptable, given visibility requirements.

A potential alternative would be using a custom view for action bar.

Prior view:

Screen Shot 2019-06-24 at 4 11 25 PM

New view, long name:

Screen Shot 2019-06-24 at 4 10 13 PM

New view, short name:

Screen Shot 2019-06-24 at 4 09 43 PM

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
+++ b/app/src/main/java/org/projectbuendia/client/ui/chart/PatientChartActivity.java
@@ -560,7 +560,12 @@ public final class PatientChartActivity extends BaseLoggedInActivity {
             labels.add(patient.birthdate == null ? "age unknown"
                 : Utils.birthdateToAge(patient.birthdate, getResources())); // TODO/i18n
             String sexAge = Joiner.on(", ").join(labels);
-            PatientChartActivity.this.setTitle(id + ". " + fullName + SEPARATOR_DOT + sexAge);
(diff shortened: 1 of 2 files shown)


ivangayton:
Hi Matt,

This is a possible resolution to the problem in [#325](https://github.com/projectbuendia/client/issues/325), and it's lovely pull request; thanks for the great notes and screenshots!

 It's a 100% improvement in the case of overflow; the text may be small but it's present. It's no smaller than some of the other text on the screen, such as the dates on the chart, though it's maybe a bit harder to see due to lesser contrast.

I'm inclined to think it's an improvement over what we currently have, and is thus worth merging simply because if we never get around to fixing #325, small text is better than no text. Nothing stopping us from considering other solutions later if need be.

I'll leave it to Ping to make the final decision, but my vote is to merge this, close #325, and see if anyone complains that the age and sex are too small to see easily.

mjanes:
Have adjusted style slightly. New appearance is easier to read.

Screen Shot 2019-06-25 at 11 33 56 AM

zestyping (approved):
Great, thanks!

Issue #330, opened 2019-06-24
Patient Record Ebola Test -> says (in the future) when test was previously recorded


saratheneale:
Edit: Confirmed this happens if the test was in the past. 

1. Admit new patient 
2. Day 2 - get an ebola test. 
3. On Day 4, observe Patient record 

Result: The Ebola test says `Ebola test (in the future)` 
Expected: `Ebola test (x days ago)`

![20190621_054430](https://user-images.githubusercontent.com/3619338/60057195-7c881800-9698-11e9-9231-5f069dd4eb8a.jpg)

PR #333, opened 2019-06-25
Fix issue 330

mjanes:
Patient Record Ebola Test -> says (in the future) when test was previously recorded

https://github.com/projectbuendia/client/issues/330

To fix, use today as anchor, and display test time relative to today.


zestyping (approved):

zestyping:
Awesome!  Thank you!

Issue #342, opened 2019-06-25
Arrange observation detail dialog fields in a sane order


zestyping:
As background, please read the wiki page [Profiles](https://github.com/projectbuendia/buendia/wiki/Profiles) for details on the layout of the CSV file and the chart screen.

When you tap a cell, column heading, or row heading on the patient chart grid, a dialog box pops up showing you a list of specific observations and their timestamps.  Look for `showObservationsDialog` to see how this happens.

The list of observations is currently in an apparently random order.   Instead, the fields should be shown in the same order as the tiles, followed by the order of the rows in the table.  Observations that aren't among the rows in the table should go in a section at the end.

zestyping:
Closed by #344.

PR #344, opened 2019-06-25
Sort obs rows in same order as chart concept uuids

mjanes:
https://github.com/projectbuendia/client/issues/342

### DIFF
--- a/app/src/main/java/org/projectbuendia/client/models/ObsRow.java
+++ b/app/src/main/java/org/projectbuendia/client/models/ObsRow.java
@@ -15,15 +15,19 @@ public class ObsRow implements Parcelable {
     public String day;
     public String time;
     public String conceptName;
+    public String conceptUuid;
     public @Nullable String value;
     public @Nullable String valueName;
 
     public ObsRow(String Uuid,
                long Millis,
-               String ConceptName,
+               String conceptName,
+               String conceptUuid,
                @Nullable String Value,
                @Nullable String ValueName) {
-        conceptName = ConceptName;
+
+        this.conceptName = conceptName;
+        this.conceptUuid = conceptUuid;
 
         DateTimeFormatter fmtDay = DateTimeFormat.forPattern("dd MMM yyyy");
         day =  fmtDay.print(new DateTime(Millis));
@@ -53,6 +57,7 @@ public class ObsRow implements Parcelable {
         day = in.readString();
         time = in.readString();
         conceptName = in.readString();
+        conceptUuid = in.readString();
         value = in.readString();
         valueName = in.readString();
     }
@@ -65,6 +70,7 @@ public class ObsRow implements Parcelable {
         dest.writeString(day);
         dest.writeString(time);
         dest.writeString(conceptName);
+        dest.writeString(conceptUuid);
         dest.writeString(value);
         dest.writeString(valueName);
     }
--- a/app/src/main/java/org/projectbuendia/client/utils/Utils.java
+++ b/app/src/main/java/org/projectbuendia/client/utils/Utils.java
@@ -31,6 +31,7 @@ import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 import org.projectbuendia.client.App;
 import org.projectbuendia.client.R;
+import org.projectbuendia.client.models.ObsRow;
 import org.projectbuendia.client.net.Server;
 
 import java.io.PrintWriter;
@@ -42,6 +43,7 @@ import java.net.URLEncoder;
 import java.text.Normalizer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/projectbuendia/client/models/ObsRow.java
```diff
@@ -15,15 +15,19 @@
     public String day;
     public String time;
     public String conceptName;
+    public String conceptUid;
```
zestyping: This should be spelled `conceptUuid`—could you rename both the field and the constructor argument?

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -50,7 +52,9 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-/** Utility methods. */
+/**
```
zestyping: Could you undo this reformat?  Comments that can fit on one line should stay on one line.

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -79,36 +84,65 @@
...
+            };
+
+
+    /**
+     * Sorts an ArrayList of ObsRow rows so that the order of the ObsRow.conceptUids matches the
+     * order of conceptRowUuids.
+     */
+    public static ArrayList<ObsRow> sortObsRows(final ArrayList<ObsRow> rows,
+                                                final ArrayList<String> conceptUuids) {
+
```
zestyping: Remove unnecessary blank lines.

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -79,36 +84,65 @@
...
+        Collections.sort(rows, new Comparator<ObsRow>() {
+                    @Override
```
zestyping: This is one level of indentation, so it should be 4 spaces in; it looks like it's 12.

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -79,36 +84,65 @@
...
+                    public int compare(ObsRow obsRow1, ObsRow obsRow2) {
+                        int indexObsRow1 = conceptUuids.indexOf(obsRow1.conceptUid);
+                        int indexObsRow2 = conceptUuids.indexOf(obsRow2.conceptUid);
```
zestyping: Avoid using names that are mostly similar and hard to tell apart, e.g. in `indexObsRow1`, the `indexObsRow` part conveys almost no information whereas the `1` is critical, so you have to visually find a needle in a haystack to detect a typo.

Although the common advice to avoid single-letter names is often correct, comparison methods are a special case because there are usually two of everything.  In this situation,  `a` and `b` are easier to tell apart than `obsRow1` and `obsRow2`:

    public int compare(ObsRow a, ObsRow b) {
        int ai = conceptUuids.indexOf(a.conceptUuid);
        int bi = conceptUuids.indexOf(b.conceptUuid);
        ... etc.
    }

app/src/main/java/org/projectbuendia/client/utils/Utils.java
zestyping: Actually I'm seeing now that your IDE must have reformatted most of this file.  Please revert the reformatting-only changes because they make it harder to see what you meant to change.  It looks like you only meant to add the one static method `sortObsRows`, I think?

For future reference, the [style conventions in this codebase](https://github.com/projectbuendia/buendia/wiki/Java-Style) generally prioritize efficient use of screen space to counterbalance Java's verbosity.  For example, we don't use line breaks after `@Override`, and comments that fit on one line should be kept on one line.

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -79,36 +84,65 @@
...
+                        if (indexObsRow1 == -1) {
+                            indexObsRow1 += rows.size();
```
zestyping: It looks like this means any row that doesn't appear in the `conceptUuids` list will get an index of `rows.size() - 1`.  I see two possible concerns with this:
  - If there is one row for every concept, then `rows.size() - 1` is also the index of the last row in the list, which would mean that it becomes indistinguishable (in the ordering) from the last item, and might end up sorted before or after the last item.  Or basically, the `+=` isn't needed because we don't care about the `-1`; we just want it to be something bigger than any other matching row.  For example, set it to `conceptUuids.size()`.
  - The non-matching items all end up with the same index, so they will get sorted into a random-looking order.  There isn't really a great way to standardize on an order to put them in, but let's make it predictable, at least, by doing a secondary comparison of the concept names when the indexes are equal.

zestyping:
Hey @mjanes!  Thanks for your work on this!  This looks pretty good; there are just a couple of things in the review comments I'd like you to address and then I'll merge this in.

app/src/main/java/org/projectbuendia/client/models/ObsRow.java
mjanes: Done

app/src/main/java/org/projectbuendia/client/utils/Utils.java
mjanes: Pardon the auto-format stomping. Is there a code styles file to import?

app/src/main/java/org/projectbuendia/client/utils/Utils.java
mjanes: Would you prefer a secondary sort by name? Or maintaining the order in the input list?

zestyping: Secondary sort by name would be better—I don't think the order in the input list is useful.

zestyping (approved):
Looks great!

zestyping:
This Closes #342.

PR #348, opened 2019-06-26
Smaller icons for children

mjanes:
Screenshot with new icons:

Screen Shot 2019-06-26 at 10 27 43 AM


zestyping:
Thanks for working on this!  Would you be able to post a screenshot here showing the new icons in context?

app/src/main/java/org/projectbuendia/client/utils/Utils.java
```diff
@@ -408,6 +408,17 @@ public static DateTime getDateTime(Bundle bundle, String key) {
         return dateTime == null ? null : MEDIUM_DATETIME_FORMATTER.print(dateTime);
     }
 
+    /**
+     * @return True if birthdate is non-null and less than or equal to five years old
+     */
+    public static boolean isChild(LocalDate birthdate) {
+        if (birthdate != null){
+            Period age = new Period(birthdate, LocalDate.now());
+            return age.getYears() <= 5;
```
zestyping, line 417: This will return true for an age like 5 years + 11 months.  Because the fix is so tiny (`<` instead of `<=`) I'll just merge this and then make the fix.

zestyping (approved):

Issue #298, opened 2019-06-18
Crash in patient list activity


zestyping:
To be investigated.

```
06-17 17:37:53.861  6467  6467 E AndroidRuntime: java.lang.NullPointerException: Attempt to read from null array
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at org.projectbuendia.client.ui.PatientListTypedCursorAdapter.getGroup(PatientListTypedCursorAdapter.java:120)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at org.projectbuendia.client.ui.PatientListTypedCursorAdapter.getChildrenCount(PatientListTypedCursorAdapter.java:124)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.widget.ExpandableListConnector.refreshExpGroupMetadataList(ExpandableListConnector.java:563)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.widget.ExpandableListConnector.expandGroup(ExpandableListConnector.java:688)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.widget.ExpandableListView.expandGroup(ExpandableListView.java:750)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.widget.ExpandableListView.expandGroup(ExpandableListView.java:734)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at org.projectbuendia.client.widgets.GroupedListView.expandAllGroups(GroupedListView.java:70)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at org.projectbuendia.client.widgets.GroupedListView.access$000(GroupedListView.java:31)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at org.projectbuendia.client.widgets.GroupedListView$2.onChanged(GroupedListView.java:63)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.database.DataSetObservable.notifyChanged(DataSetObservable.java:37)
06-17 17:37:53.861  6467  6467 E AndroidRuntime: 	at android.widget.BaseExpandableListAdapter.notifyDataSetChanged(BaseExpandableListAdapter.java:56)
[...]

PR #350, opened 2019-06-26
Bandaid for issue 298

mjanes:
https://github.com/projectbuendia/client/issues/298


zestyping:
This is marked draft so I can't merge it, but it seems harmless and probably helpful so I'll make this change in `dev`.

PR #354, opened 2019-06-27
Improve back navigation by upgrading to Toolbar. Closes #336

### LINKED ISSUE #336: Make back navigation more consistent
Navigating back to the previous activity is currently done in many different ways:
  - Sometimes the app icon in the top-left corner is a Back button
  - Sometimes the app icon in the top-left corner _and_ the text next to it is a Back button
  - Sometimes the app icon in the top-left corner doesn't do anything
  - Sometimes you have to open the user menu and tap "Logout" to go back

This is confusing.  Let's consolidate this behaviour as follows:
  - On the Login screen:
      - The icon in the top-left corner is the app icon.
      - There is no way to go Back.  (Tapping the system Back button does not quit the app.)
  - On all screens except the Login screen:
      - The icon in the top-left corner is always a left-pointing arrow.
      - There's no tiny chevron in the top-left corner because the entire square icon is an arrow.
      - Tapping the left-arrow icon in the corner (and never the text next to it) goes back.
      - Pressing the system Back button goes back.
      - Tapping the left-arrow icon and tapping the system Back button always does the same thing.

The activity stack should always be a prefix of the following sequence:

Login > Location List > Patient List > Patient Chart > Form

Navigating forward and pressing Back should always step right or left in this chain.

When the GoToPatientDialog is used to jump to a patient, the new patient chart is not placed on the activity stack on top of the current activity; instead it replaces the current activity.  Stepping Back from the Patient Chart should go to the Patient List for the location where that patient is located.

krtonga:
Issues: Closes #336 
Scope: All top menu navigation
Requested reviewers: 

#### User-visible changes

The `appbar` has been replaced with the newer `toolbar`, which looks and behaves as expected. 

The chart and patient list activities have been made to only exist once in the stack, resulting in a back button experience that is much simpler.

_Two known minor bugs:_ 
1. The toolbar properly increases in size with the `+` and `-` volume buttons, but the signed in user icon resizes too fast, and the title font size is not increasing. 
2. The settings activity uses a different theme from the rest of the app.

#### Internal changes 

Changing to toolbar required me to use `AppCompatActivity`, and use a `NoActionBar` theme. This resulted in some deprecations where we use toolbar tabs, which will need to be examined at a later point.

Also to note: Because settings screen was implemented using `PreferenceActivity` which inherits from `Activity`, I added an extra class that uses a delegate to get `AppCompat` functionality. Instead of using a custom layout, I simply use a default `AppCompat` theme with action bar.

#### Guidance for testers 

Please double check all app bar navigation.

I realize that this upgrade added time and complexity to this PR. I attempted some hacky solutions to make the back button look and feel as expected, and was not getting the results I wanted. The [Google Developers documentation strongly recommended using `toolbar` instead](https://developer.android.com/training/appbar/setting-up.html).

Also to note: I attempted a few things on the `SettingsActivity` including a custom layout and using a custom theme. I kept running into bugs, and I decided it was not worth the time investment at this junction.

#### Screenshots 
336-toolbar
336-zoom-bug
336-toolbar-settings

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -95,7 +95,8 @@
             android:name=".ui.lists.FilteredPatientListActivity"
             android:label="@string/title_patient_list"
             android:parentActivityName=".ui.lists.LocationListActivity"
-            android:screenOrientation="userPortrait" >
+            android:screenOrientation="userPortrait"
+            android:launchMode="singleInstance">
             <intent-filter>
                 <action android:name="android.nfc.action.TAG_DISCOVERED" />
                 <category android:name="android.intent.category.DEFAULT" />
@@ -106,7 +107,8 @@
             android:label="@string/title_patient_chart"
             android:parentActivityName=".ui.lists.FilteredPatientListActivity"
             android:screenOrientation="userPortrait"
-            android:uiOptions="splitActionBarWhenNarrow">
+            android:uiOptions="splitActionBarWhenNarrow"
+            android:launchMode="singleInstance">
             <meta-data
                 android:name="android.support.UI_OPTIONS"
                 android:value="splitActionBarWhenNarrow" />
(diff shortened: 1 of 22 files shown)


zestyping:
Thanks for figuring this out!  I went with your other branch, `krtonga/back-buttons`.  It was rebased and merged in #382.

PR #356, opened 2019-06-28
Fix or disable all remaining failing tests for a green test run

sdspikes:

#### Internal changes

We can now add tests to circleCI integration and have failures be indicative of actual problems.

#### Verification performed

Ran the tests :)

#### Guidance for reviewers

There are several tests that work if you set a breakpoint and wait long enough (or put in a long sleep) but I decided to disable them because I didn't want the tests to take forever.  I started looking into IdlingResource stuff but decided to just get to a green test run first and start working on new tests rather than sink too much time into it right now.


sdspikes:
Note for circleCI setup: android tests can be run from the command line with

`./gradlew connectedAndroidTest`

However, it requires an emulator to be running and connected.

zestyping (approved):
Hurrah!  I got a few failures (4 out of 165) when I ran these tests, but they were all easily explainable.  3 were due to there being too many user accounts already on the server (so the newly created account wasn't visible, or there were multiple accounts with initials "GU"), and the remaining failure is easy to fix.  This is hugely helpful!

Issue #297, opened 2019-06-17
Enhancement: Treatments design with adaptive individual columns


ivangayton:
Here's the idea of the treatments section of the chart.

The principle is:

- Day columns always align with the days in the chart
- Within a day, each row in Treatments has its own individual number of columns, and therefore width of subdivided columns. 

![Treatments Design](https://user-images.githubusercontent.com/5991943/59711808-a9cc5600-91c0-11e9-9848-2769056c8ad1.png)

And the proposed entry form for treatments:
![Treatments entry form](https://user-images.githubusercontent.com/5991943/59711816-af29a080-91c0-11e9-8218-b4b6fddbebe3.png)

zestyping:
The main work of the chart display is now done (at least to a prototype level) in #372.  Closing this and opening new issues #375 and #376 for the remaining smaller changes to the treatment order dialog.

PR #372, opened 2019-07-01
Represent and display scheduled doses as intervals

zestyping:
Issues: Progress toward #297.
Scope: chart treatment display, order dialog

#### User-visible changes

The doses scheduled by each order are shown as intervals in the Treatments section of the chart; the interval boundaries need not match column boundaries.  Within each interval, the number of administered doses and number of scheduled doses are shown, with colours to indicate underdosing, overdosing, and doses that are due:

  - Underdose, in the past: amber
  - Underdose, in the present (i.e. due for administration): red
  - Fully dosed, in the past or present: green
  - Overdose, in the past or present: purple

#### Internal changes

There are two types of orders: unary orders and series orders.  Unary orders represent a single dose scheduled at a point in time, and have no frequency.  Series orders represent a series of doses scheduled at regular intervals from a start time until a stop time, and have a frequency specified as an integer number of doses per day.

Each dose is modelled as a "dose interval"; a dose is considered to be given "on time" if the actual time of administration falls within the dose interval.  A unary order has one dose interval, from the prescribed time until the end of that day.  A series order has a series of dose intervals, formed by dividing each day into equal divisions according to its frequency, and consisting of all the divisions that contain or come after the start time but do not contain or come after the stop time.  Note that the divisions (and hence the dose intervals) for a given order are not necessarily of uniform length, because certain days are longer or shorter than 24 hours.

#### Verification performed

I created and edited several orders of varying durations and frequencies, and viewed them at various column zoom levels.

#### Screenshots

Cells are coloured according to dosing status:
order-intervals

When multiple intervals are packed into a single column, the colours show the individual intervals, while the text summarizes the number of doses given and scheduled for the entire duration of the column:
split-order-cells

Columns can be divided into arbitrarily many intervals; overdoses are shown in purple with red numbers:
overdose


zestyping:
@ivangayton Visual design shown in screenshots for your perusal.

PR #374, opened 2019-07-01
Add route field to order dialog; store instructions using field separators; fix interpretation of orders with no stop time.

zestyping:
Issues: Progress toward #297.
Scope: treatments section of patient chart, order dialog

#### User-visible changes

The order dialog now offers a selection for the route of administration (PO, IV, IM, or SC).

Orders that have no frequency are considered to be administered exactly once (on the day of the order).

Orders with a frequency but no stop time are considered to be administered indefinitely.

#### Internal changes

The instruction fields (medication, route, dosage, frequency) are now packed into the OpenMRS "instructions" field using ASCII record separator characters (30 and 31) instead of fiddly regular expressions.

#### Screenshots

route-of-administration

### DIFF
--- a/app/src/main/assets/chart.html
+++ b/app/src/main/assets/chart.html
@@ -139,7 +139,7 @@
       {% for order in orders %}
         <tr class="order">
           <th scope="row" onclick="controller.onOrderHeadingPressed('{{order.uuid}}')">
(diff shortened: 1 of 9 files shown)


zestyping:
@ivangayton More progress on treatments.

PR #383, opened 2019-07-02
Adds a "Notes" field to treatment orders.

### LINKED ISSUE #337: Treatment order: Add a free-text notes field
Treatment orders should have a free-text field for additional notes or directions.

 - The treatment order dialog should have an additional text entry field.
 - The `Order` model needs to be able to store this field.
 - The local database schema for the `orders` table should be updated to add this field.
 - The JSON object and sync task for orders needs to store and retrieve this field from the server.
 - The treatment order display in the chart should show an appropriate icon (yellow caution triangle?) and/or the first few characters/words of the note in the leftmost cell (the row header).

zestyping:
Issues: Closes #337 
Scope: treatment order dialog, order rows in grid

### DIFF
--- a/app/src/main/assets/chart.css
+++ b/app/src/main/assets/chart.css
@@ -110,7 +110,15 @@ th, td {
   padding-bottom: 3.8rem;
   font-weight: bold;
 }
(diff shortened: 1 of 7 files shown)


zestyping:
@ivangayton  Here's what this looks like:

order-dialog-with-notes

order-with-notes-preview

PR #115, opened 2015-11-04
Add a localhost server to the build.gradle

capnfabs:

PR #393, opened 2019-07-10
Sign release APKs in CircleCI using 'zestybuendia' keystore.

schuyler:
Scope: Add CircleCI configuration, along with changes to Gradle, to enable building release APKs from a consistent Android developer keystore.

#### User-visible changes

Users will be able to upgrade from this point forward without an "app not found" error.

#### Internal changes

This CircleCI configuration requires the presence of `ANDROID_KEYSTORE` and `ANDROID_KEYSTORE_PASSWORD` variables set in the [project environment](https://circleci.com/gh/projectbuendia/client/edit#env-vars).

The `ANDROID_KEYSTORE` env variable was set by running `base64 < zestybuendia.jks | pbcopy` on a local MacOS laptop, and then pasting the results into the CircleCI config UI.

#### Verification performed

This configuration was tested on CircleCI in [build #115](https://circleci.com/gh/projectbuendia/client/115#artifacts/containers/0).

#### Screenshots

Screen Shot 2019-07-09 at 19 40 33

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -232,11 +232,20 @@ android {
     }
 
     signingConfigs {
-        release {
-            storeFile file(new File(releaseRepoRoot, 'keys/org.projectbuendia.client.jks'))
-            storePassword ''
-            keyAlias 'org.projectbuendia.client'
-            keyPassword ''
+        if (System.getenv("CI")) {
+            release {
+                keyAlias 'buendia'
+                keyPassword System.getenv('ANDROID_KEYSTORE_PASSWORD')
+                storeFile file(System.getenv('ANDROID_KEYSTORE_FILE'))
+                storePassword System.getenv('ANDROID_KEYSTORE_PASSWORD')
+            }
+        } else {
+            release {
+                storeFile file(new File(releaseRepoRoot, 'keys/org.projectbuendia.client.jks'))
+                storePassword ''
+                keyAlias 'org.projectbuendia.client'
+                keyPassword ''
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -277,14 +286,20 @@ gradle.taskGraph.whenReady {
             if (!android.signingConfigs.release.storeFile.exists()) {
                 throw new IllegalStateException("To perform a release build, you must have the release repo cloned at ${releaseRepoRoot}.")
```
zestyping: This error message is out of date—it should say something like "there is no keystore file at [path]".

app/build.gradle
```diff
@@ -277,14 +286,20 @@ gradle.taskGraph.whenReady {
...
             }
-            if (System.console() == null) {
+
+            if (System.console() == null && !System.getenv("CI")) {
```
zestyping: This check is here just to ensure that we can read the password from the console, so it should get moved into the `if` block below (and the message could be massaged a bit).

app/build.gradle
```diff
@@ -232,11 +232,20 @@ android {
...
-            keyPassword ''
+        if (System.getenv("CI")) {
+            release {
+                keyAlias 'buendia'
+                keyPassword System.getenv('ANDROID_KEYSTORE_PASSWORD')
+                storeFile file(System.getenv('ANDROID_KEYSTORE_FILE'))
+                storePassword System.getenv('ANDROID_KEYSTORE_PASSWORD')
+            }
+        } else {
+            release {
```
zestyping, line 243: There's no "release" repo anymore.  Let's change the path to a local location for `zestybuendia.jks` (maybe the root of the `client` repo or nearby?) and use the same `keyAlias` as above so that we can build compatible APKs locally.

zestyping (approved):
Feel free to merge after addressing my comments.  Thanks!

app/build.gradle
schuyler: Fixed

app/build.gradle
schuyler: Fixed

app/build.gradle
schuyler: Fixed

PR #402, opened 2019-07-11
Add version numbering to dev builds

### LINKED ISSUE #401: Release APKs lack versionCode
I've just downloaded the APK from the [last release](https://github.com/projectbuendia/client/releases/tag/v0.12), and it lacks the value for `versionCode`:

    package: name='org.projectbuendia.client.dev' versionCode='' versionName='7af6a39' platformBuildVersionName='7af6a39'

As `versionCode` is what Android uses to tell versions apart (and to decide which is newer), this is an important technical detail. Could you please ensure it is set? And while on it, the commit hash is a bad idea for `versionName` – which is intended to be human-readable, so humans can tell versions apart; especially for non-techies it is hard to tell which hash is newer, so while on it, could you "sync" that with the tag name?

Thanks in advance!

schuyler:
**NOTE**: This PR was doing too much and so I've split the CircleCI refactoring out into #406. This PR should be _not_ be merged until after that one. 

### Issues:

Closes #401.

### Scope:

* Add APK version numbering for dev builds based on CircleCI build number.

### User-visible changes

Both dev and release builds will have a version code and a human-readable version string.

Builds made from the command line will DTRT if the developer runs `./gradlew -DversionNumber=$(.circleci/get_package_version) assembleDebug`.

### DIFF
--- a/.circleci/get_package_version
+++ b/.circleci/get_package_version
@@ -30,17 +30,18 @@ if echo "$CIRCLE_TAG" | grep -Eq '^v([0-9]+\.?)+$'; then
 fi
 
 # Find the most recent release tag on master.
-TAG=$(git describe --match='v*' --abbrev=0 origin/master)
+TAG=$(git describe --tags --match='v*' --abbrev=0 origin/master)
 
-# If we're building in CircleCI, the version is "x.y.z+b<n>" where <n> is the
+# If we're building in CircleCI, the version is "x.y.<b>" where <b> is the
 # latest build number. 
 if [ -n "$CIRCLE_BUILD_NUM" ]; then
-    echo "${TAG#v}+b${CIRCLE_BUILD_NUM}"
+    IFS='.' read -ra TAG_VERSION <<< "${TAG#v}"
+    echo "${TAG_VERSION[0]}.${TAG_VERSION[1]}.${CIRCLE_BUILD_NUM}"
     exit
 fi
 
-# Otherwise, we're building locally, so the version should be "x.y.z~g<commit>"
-# where <commit> is the latest unambiguous commit identifier. This version is
-# actually treated by Debian as _less_ recent than "x.y.z".
+# Otherwise, we're building locally, so the version should be "x.y.z-<commit>"
+# where <commit> is the latest unambiguous commit identifier. "+dirty" is
+# appended if the current tree is dirty.
 COMMIT=$(git describe --abbrev --dirty=+dirty)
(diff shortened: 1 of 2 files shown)


.circleci/get_package_version
```diff
@@ -30,17 +30,18 @@ if echo "$CIRCLE_TAG" | grep -Eq '^v([0-9]+\.?)+$'; then
 fi
 
 # Find the most recent release tag on master.
-TAG=$(git describe --match='v*' --abbrev=0 origin/master)
+TAG=$(git describe --tags --match='v*' --abbrev=0 origin/master)
 
-# If we're building in CircleCI, the version is "x.y.z+b<n>" where <n> is the
+# If we're building in CircleCI, the version is "x.y.<b>" where <b> is the
```
zestyping, line 35: Why omit `z` here?

schuyler: Our `app/build.gradle` expects `versionNumber` to be in the form _\<major>.\<minor>.\<patch>_. If we want the build number to appear in both the `versionString` and `versionCode`, it has to go somewhere, and I figured replacing the _\<patch>_ number of the latest release was the most innocuous way to do that.

The alternative would be to continue to append the build number to the original `versionNumber` but then (a.) `build.gradle` would need to be updated to permit that and (b.) the `versionCode` wouldn't change between dev builds. (Maybe that's actually what we want?)

zestyping: First, just to be clear on terminology, the fields are named `versionName` and `versionCode`.

The only practical effect of the `versionCode` is to prevent unintended downgrades.  The Play Store will only perform automatic updates that increase the `versionCode`.  As far as I know, only the Play Store enforces this; the phone doesn't care whether the `vesionCode` increases or not. Because we don't plan to ever publish development APKs to the Play Store, in some sense it doesn't matter what we use for dev builds.

Any of these choices are fine:
  - Set `versionCode` to 0 for all dev builds
  - When `versionName` is `x.y.z`, set `versionCode` to 1000000x + 10000y + z for all dev builds
  - When `versionName` is `x.y.z` at patchlevel p, set `versionCode` to 1000000x + 10000y + 100z + p for all dev builds
  - When `versionName` is `x.y.z`, set `versionCode` to 1000000x + 10000y + p for all dev builds

The first is the simplest (it's what `build.gradle` currently does, and I think it will work fine—the net effect is that any production release will supersede (automatically install over) any development release.

PR #406, opened 2019-07-11
DRY out the CircleCI configuration

schuyler:
This PR has been split out from #402 for simplicity and clarity.

* Refactor CircleCI configuration using YAML references to eliminate repeated stanzas. Note that this is inspired by what [ODK Collect](https://github.com/opendatakit/collect/blob/master/.circleci/config.yml) is doing.
* Bump CircleCI build image to Android API 28, since that's what Gradle is using.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,99 +1,121 @@
 version: 2
-jobs:
-  build:
+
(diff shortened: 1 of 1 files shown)


zestyping (approved):
Sorry it took me a while to get to this.  I understand what's happening here now and it seems fine.

PR #408, opened 2019-07-14
Autocompletion for medications

### LINKED ISSUE #341: Drop-down list of 10 most common medications

zestyping:
Issues: Closes #341.
Scope: Treatment order dialog

#### User-visible changes

The Medication field in the treatment order dialog now performs autocompletion.  Possible completions appear once the first letter has been typed, and narrow down as more letters are typed.

The list of known medications comes from MSF's Essential Drugs guide, 2016 edition.  I also added a few additional Ebola treatments and vaccines to the list.

The typed letters are matched against the beginnings of words in the medication name, including alternative names for the medication (shown in parentheses).  They can match the medication name with punctuation treated as a word break or with punctuation removed; for example, "Co-artemether" can be made to appear in the completion list by typing "artem" (treating the hyphen as a word break) or "coart" (treating the hyphen as nonexistent).

#### Verification performed

I opened the dialog and tried typing in the names of drugs in several ways.

1. Originally, we considered showing a list of the 10 most common medications, where this list would be configured by the MTL.  I'm opting for an autocompletion approach instead, because (a) autocompletion speeds up the entry of a much larger range of possible medications, and (b) using a comprehensive list of medications makes it more reasonable to build a fixed list into the app and avoid the complexity of providing a way for the MTL to edit the list.

2. The typed text is matched against the beginning of each word, not as a substring anywhere within the name.  This is an intentional choice, made so that typing a single letter will often be enough to bring up the desired medication.  For example, typing "I" brings up Ibuprofen immediately.  If we used substring matching, "I" would match over 100 items, with Ibuprofen buried somewhere far down the list.  It also seems more intuitive that typing "I" should bring up the things that start with "I", not everything that contains a "I".

Further improvements are under consideration:
  - The drop-down list currently extends down below into the keyboard area.  The keyboard obscures that part of the list, so you cannot reach the last 3 or 4 entries.  The list should be limited to the area above the keyboard so that you can scroll all the way to the end of the list if necessary.
  - Instead of showing the matches in alphabetical order, we should give more weight to the ones that have been used recently.
  - The drop-down list doesn't open immediately when you open the treatment order dialog; you still have to type the first letter.  We might want to make it open immediately if the entries are sorted by popularity.

#### Screenshots

type-p

type-pa


zestyping:
Medications are now shown in the list with a caption giving the general category or purpose of the medication.

The drop-down list is now sized so that it doesn't get covered up by the soft keyboard.

completion-with-captions

PR #423, opened 2019-08-07
Fix up how location names are displayed

### LINKED ISSUE #414: "All present patients" title looks like a spinner but doesn't do anything
App version: 0.15
Screen or dialog: Patient list

zestyping:
Issues: Closes #414.

Location names may be prefixed with extra information in square brackets.  This information is not displayed on the client, but:

  - It is considered when sorting; thus you can control the sort order of locations by putting "[1]", "[2]", etc. in front.  The sort is alphanumeric, so "[2]" will properly come before "[11]".
  - It allows you to specify the default location with an asterisk, e.g. "[2*] Triage"


zestyping:
This also closes #414.

Issue #400, opened 2019-07-10
Sync: Changing Patient Location doesn't Update other Devices with that Screen


saratheneale:
App version: Buendia 0.14
Screen or dialog: Patient Chart

#### Steps to replicate the problem
Have 2 devices each on the Patient Chart screen. Device 1 will Edit. Device 2 will observe.

1. On Device 1, edit Patient Location from Triage to S2
2. On Device 2, wait 5 minutes.
Result: The patient location still says Triage. 

#### Actual behaviour
The patient location still says Triage. 
Notes: 
Navigating back to Patient List and Clicking back into Patient still does not show the updated Location. Neither does adding a Round. The only thing that updates it is manually refreshing the Patient List. 

#### Expected behaviour
Result: The patient location should say S1 within 1 minute.

zestyping:
Closed by #430.

PR #430, opened 2019-08-07
Improve efficiency of forest updating based on patient changes

zestyping:
With this PR, when patients are moved from one location to another, the updating of patient counts on the location screens is smooth and pretty responsive.


zestyping:
I've done some testing with multiple devices, and I'm pretty sure this closes #400.

PR #488, opened 2019-11-23
Add a "None of the above" button to multiple-select groups.

### LINKED ISSUE #367: Observation -> Can't null out or false symptoms
App version: (the number after "Buendia" on the starting screen)
Screen or dialog:

#### Steps to replicate the problem

1. Go to Patient Observation. Click New Observation
2. Click Nosebleed.
3. Save
- [ ] Confirm Nosebleed is in observation chart. 
4. Click New Observation
5. Try to remove the nosebleed observation. It's not obvious
6. Click another symptom
7. Save. 

#### Actual behaviour
You can't remove the nosebleed

#### Expected behaviour
There should be a way to remove the nosebleed symptom in the round.

zestyping:
Issues: Closes #367.
Scope: BinarySelectOneWidget and its table group builder

#### User-visible changes

In each multiple-select question:

1. Previously each option would be recorded with a "yes" or a "no" value; the default value is "no" and tapping an option toggles between "yes" and "no".  Thus, submitting the form without doing anything would record a "no" for all the options.  Now, the default value is "unanswered" and tapping an option toggles between "yes" and "unanswered", so that submitting the form without doing anything doesn't record anything for such questions.

2. There is a new "None of the above" button at the bottom.  Tapping this button sets the state of all the options to "no".  This button is the only way to set options to "no" (rather than "unanswered").  When "None of the above" is selected, tapping any other option clears the "None of the above" button and reverts all the other options to "unanswered".

The appearance of options changes slightly:

  - "Yes" state: Previously, black text on green background.  Unchanged.
  - "No" state: Previously, black text on white background.  Now, grey text on white background.
  - "Unanswered" state: Previously, black text on white background.  Unchanged.

#### Internal changes 

The unwieldy name "BinarySelectOneTableWidgetGroupBuilder" (which ought to win some sort of prize) has been renamed to the more concise "BinarySelectTableBuilder".

#### Verification performed

I tested this manually on a tablet and confirmed that the options light up and clear as expected, in response to taps on the option buttons and the "None of the above" button.  I submitted the form and confirmed that nothing is recorded if nothing is tapped, and that "no" is recorded for all options if "None of the above" is selected.

This design was intended to meet the following goals:

  - We don't want to record "no" for a symptom when the patient hasn't been examined at all for that symptom.
  - We do want a way for medics to record "no" for all symptoms when they've examined the patient.
  - We don't want the UX to be too complicated.

#### Screenshots 

Here is a short video of the interaction:

[none-of-the-above-mp4.zip](https://github.com/projectbuendia/client/files/3882186/none-of-the-above-mp4.zip)

Here's what it looks like with a couple of options selected:

none-of-the-above-with-selections

And here it is with the "None of the above" button selected:

none-of-the-above-none-selected

### DIFF
--- a/third_party/odkcollect/src/main/java/org/odk/collect/android/widgets2/group/BinarySelectOneTableWidgetGroupBuilder.java
+++ b/third_party/odkcollect/src/main/java/org/odk/collect/android/widgets2/group/BinarySelectOneTableWidgetGroupBuilder.java
@@ -116,9 +116,9 @@ public class BinarySelectOneTableWidgetGroupBuilder implements
         cb.setText(label);
         cb.setOnClickListener(new View.OnClickListener() {
             @Override public void onClick(View v) {
-                ((CheckBox) v).setChecked(true);
+                Boolean newState = ((CheckBox) v).isChecked() ? false : null;
                 for (BinarySelectOneWidget widget : mWidgets) {
-                    widget.setState(false);
+                    widget.setState(newState);
                 }
             }
         });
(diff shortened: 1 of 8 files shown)


zestyping:
Ivan has tried this out and it looks good!

